PR #8, opened 2019-11-07
Fix: installation instructions

minthemiddle:
- Add: Laravel required
- Fix: Blueprint tries to load `draft.yml`
- Add: Copy sample yml

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,13 @@ _**v0.1 Tagged**: A beta release of Blueprint is now available which supports ge
 
 
 ## Installation
+
+To use Blueprint, you need a Laravel app:
+
+```sh
+laravel new APP_NAME
+```
+
 You can install this package via composer using this command:
 
 ```sh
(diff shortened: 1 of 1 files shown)


jasonmccreary:
Really appreciate this super quick and thorough PR.

I feel the _Laravel_ application additions are a given. They also make the installation section feel long. However, I may add a note about the Laravel 6.x requirement.

minthemiddle:
I feel this project will be a perfect starting point for beginners.
It would be a great service for them to see every step they need to follow.
More advanced users will just skip it anyway.

You might consider adding this if my impression proves right and lots of newcomer flood the issues with questions of how to use this tool.

jasonmccreary:
Fair, I'll reopen this if so.

To your point, I have an idea for a `--explain` option which would provide more output to help educate devs on the reasoning behind some of the decisions Blueprint makes when generating code. Especially when the `controllers` definitions become available.

Issue #17, opened 2019-11-15
Build out Rules translator


jasonmccreary:
While sufficient to past the current tests, the [`Rules` translator](https://github.com/laravel-shift/blueprint/blob/controller-lexer/src/Translators/Rules.php) is incomplete.

Given a `Column` definition, the `Rules` translator should be have the information it needs to intelligently translate this data into validation rules.

It can do so by:

- Column name (e.g. `email` or `email_address` adds an `email` validation rule)
- Relationship (e.g. `user_id` adds an `exists` validation rule with proper attributes)
- Data Type (e.g. `integer`)
- Attributes (e.g. `max` lengths, `digits` precision,`in` set, etc)
- Modifiers (e.g. `gt:0` for unsigned, `unique`, etc)

I expect this will constantly be tweaked. I plan to finish a base version for v1.0.

Nonetheless, I welcome PRs to begin any portion of the items above. However, large or small. I only ask that you add to the test.

To help get started, you can review the [current implementation](https://github.com/laravel-shift/blueprint/blob/controller-lexer/src/Translators/Rules.php) and review the [available validation rules](https://laravel.com/docs/5.8/validation#available-validation-rules) to get ideas of the types of things that can be translated.

jasonmccreary:
While there are likely more rules to add or boundary cases to tweak, I think this is pretty solid. I'm going to close this issue while opening #24 for future discussion.

PR #18, opened 2019-11-15
implementing nullable email rule

Treggats:
Implementing part of issue #17 

Added a test for nullable email rule.

### DIFF
--- a/src/Generators/Statements/FormRequestGenerator.php
+++ b/src/Generators/Statements/FormRequestGenerator.php
@@ -31,7 +31,9 @@ class FormRequestGenerator implements Generator
 
         $stub = $this->files->get(STUBS_PATH . '/form-request.stub');
 
-        $this->registerModels($tree['models']);
+        if (isset($tree['models'])) {
+            $this->registerModels($tree['models']);
+        }
 
         /** @var \Blueprint\Controller $controller */
         foreach ($tree['controllers'] as $controller) {
--- a/src/Translators/Rules.php
+++ b/src/Translators/Rules.php
@@ -8,8 +8,10 @@ class Rules
 {
     public static function fromColumn(Column $column)
     {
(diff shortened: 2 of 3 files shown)


jasonmccreary:
Thanks for sending this in. A few quick things preventing me from merging.

- Does it make sense for a rule to be both `required` and `nullable`? Or is it one or the other...
- There is no data type of `email`. The type would be `string` and the name of the column would be `email` or `email_address`. For clarity, the data types are any of the [migration column types](https://laravel.com/docs/5.8/migrations#creating-columns).
- Ideally, if `email` was used, it would ignore the `dataType` check. Preventing a rule from being `required|string|email`.

As you can tell, this isn't so easy. A lot of work involved for even the most basic field. So I welcome the help.

If you want to make these changes, great. Otherwise, I'll expand on this next week.

Treggats:
Just a short note, I'm going to work on this this weekend.

jasonmccreary:
No worries.

After closer reading, I think we'll disregard `nullable` for v1. If the developer has chosen to define a field to validate, the base rule should always be `required`.

Treggats:
@jasonmccreary I removed the nullable rule check, left a todo for post v1

I check if a email field is present by the name of the column and apply the rule.
The other string rules were moved to a private method with a guard clause if the email rule was applied. If that's the case the other string rules are ignored.

jasonmccreary:
Awesome. Looks like someone opened https://github.com/laravel-shift/blueprint/pull/19.

So I'll likely take the best of both PRs to mark this issue complete.

I think the last remaining areas might be `unique` and _relationship_ validations using `exists`. If you feel like tackling that that, great. Otherwise, I'll live-stream it this week.

jasonmccreary:
#19 crushed the tasks. So I'm going to close this one and pull what's needed into it.

PR #19, opened 2019-11-16
Adds various validation rules based on column name and dataType

marksalmon:
- if the dataType is string and the column name starts with 'email', the ```string``` rule gets replaced with the ```email``` rule
- if the dataType is string and the column name is 'password', the ```string``` rule gets replaced with the ```password``` rule
- If the dataType is numeric the ```numeric``` rule is added. If the dataType is unsigned, ```gt:0``` is also added
- added `in:` rule with attributes for enums and sets
- added `date` rule for dataTypes date, datetime and datetimetz

Issue #17 

Thanks JMac, Loving the streams

### DIFF
--- a/src/Translators/Rules.php
+++ b/src/Translators/Rules.php
@@ -56,6 +56,10 @@ class Rules
             $rules = array_merge($rules, ['in:' . implode(',', $column->attributes())]);
         }
 
+        if (in_array($column->dataType(), ['date', 'datetime', 'datetimetz'])) {
+            $rules = array_merge($rules, ['date']);
+        }
+
         if ($column->attributes()) {
             if (in_array($column->dataType(), ['string', 'char'])) {
                 $rules = array_merge($rules, ['max:' . implode($column->attributes())]);
--- a/src/Translators/Rules.php
+++ b/src/Translators/Rules.php
@@ -31,10 +31,6 @@ class Rules
             }
         }
 
-        // if () {
-        // $rules = array_merge($rules, ['email']);
-        // }
-
         return $rules;
     }
 
--- a/src/Translators/Rules.php
+++ b/src/Translators/Rules.php
@@ -3,6 +3,7 @@
 namespace Blueprint\Translators;
 
 use Blueprint\Column;
+use Illuminate\Support\Str;
 
 class Rules
 {
@@ -20,8 +21,9 @@ class Rules
 
         // hack for tests...
         if (in_array($column->dataType(), ['string', 'char', 'text', 'longText'])) {
-            $rules = array_merge($rules, ['string']);
+            $rules = array_merge($rules, [self::overrideStringRuleForSpecialNames($column->name())]);
         }
+
 
         if ($column->attributes()) {
             if (in_array($column->dataType(), ['string', 'char'])) {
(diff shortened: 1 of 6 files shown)


src/Translators/Rules.php
```diff
@@ -31,4 +68,19 @@ public static function fromColumn(Column $column)
...
-}
\ No newline at end of file
+
+    private static function overrideStringRuleForSpecialNames($name)
+    {
+        switch ($name) {
+            case Str::startsWith($name, 'email'):
+                return 'email';
+                break;
+            case $name === 'password':
```
jasonmccreary: This is an interesting case statement, wouldn't it just be `'password'`

src/Translators/Rules.php
```diff
@@ -20,7 +21,43 @@ public static function fromColumn(Column $column)
 
         // hack for tests...
         if (in_array($column->dataType(), ['string', 'char', 'text', 'longText'])) {
-            $rules = array_merge($rules, ['string']);
+            $rules = array_merge($rules, [self::overrideStringRuleForSpecialNames($column->name())]);
+        }
+
+        if (in_array($column->dataType(), [
```
jasonmccreary: There's probably an opportunity to get a bit more precise. For example, any of the _integer_ data type columns could use `integer` validation rule instead of `numeric`.

jasonmccreary:
This is great. I've made some inline notes of some minor tweaks. If you have time to fix great. Otherwise, I will definitely be using this as a base when I continue early this week.

marksalmon:
Hi JMac,

I've been looking at the `unique` rule and it is a little more interesting than the others because it requires the database table context. This leads me to 2 options for implementation:

1. The blueprint syntax is `validate: email:unique` and then we take the model context from blueprint (lets say 'User' for example), make sure that email is indeed a field on that model and write the rule "unique:users". This method would have to be part of the FormRequestGenerator file as the Rules class does not have any knowledge of the model context.

2. We allow the user to specify the table themselves by making the blueprint syntax 
`validate: email:unique,connection.table` and using the passed attributes to build up the rule "unique:connection.table". This method could be part of the Rules class as the model context is explicitly stated by the user in the blueprint file.

jasonmccreary:
Yes, I thought this the other day as well. I will likely refactor the `Rules` translator to take additional arguments. For now, build what you can and leave a failing test so I know to finish it out before merging.

I want Blueprint to really focus on DX. I also don't want it to mimic all of Laravel's functionality. So I am comfortable with there being a bit of work needed after generating these.

Nonetheless, we can likely get all the way there so long as the _context_ is passed in. So build what you want and use a placeholder where needed with a failing test.

marksalmon:
I've been down a major rabbit hole on this one.

It wasn't hard to get the unique rule to work by passing the context into `Rule::fromColumn()`.

The problem is that I have found that if you use a modifier after a dataType with attribute(s) you lose the attributes that are needed in the MigrationGenerator... for example....

```yaml
models:
  ModelOne:
    column: string:100 unique
```
parses out in the migration as 

```php
$table->string('column')->unique();
``` 
instead of
```php
$table->string('column', 100)->unique();
```

In the Model Lexer the attributes were overwritten with each loop through `$tokens` which means that if the first token had an attribute, and then subsequent modifiers were encountered, the original attributes were lost.

This was solved by handling the attributes for dataType and modifiers separately within the if statement for each type of token.

There is a further bug that I found during all of this:  

```yaml
models:
  ModelOne:
    column: string:100 charset:utf8
```
parses out in the migration as 

```php
$table->string('column', 100)->charset(utf8);
``` 
instead of (quotes around utf8) 
```php
$table->string('column', 100)->charset('utf8');
```

the `modelAttributesProvider` in ModelLexerTest.php suggests you want to be able to write these directives with or without quotes. At the moment the tests are all passing but only because 
`model-modifiers.bp` has the quoted attribute

jasonmccreary:
Yeah, that's kind of why I saved this for last ;)

However, looks like you crushed it dude. I will review this tomorrow in more detail and get it merged.

👍

jasonmccreary:
Not sure how to push changes directly to this PR. So I am going to merge it as is since the `controller-lexer` branch has not been merged/tagged yet anyway.

Issue #15, opened 2019-11-13
Path to draft.yaml


ManojKiranA:
I have a fresh installation of Laravel with blueprint installation after running the command

`php artisan blueprint:build`

  ```
 Illuminate\Contracts\Filesystem\FileNotFoundException  : File does not exist at path draft.yaml

  at D:\xampp\htdocs\websockets\vendor\laravel\framework\src\Illuminate\Filesystem\Filesystem.php:52
    48|         if ($this->isFile($path)) {
    49|             return $lock ? $this->sharedGet($path) : file_get_contents($path);
    50|         }
    51| 
  > 52|         throw new FileNotFoundException("File does not exist at path {$path}");
    53|     }
    54| 
    55|     /**
    56|      * Get contents of a file with shared access.

  Exception trace:

  1   Illuminate\Filesystem\Filesystem::get("draft.yaml")
      D:\xampp\htdocs\websockets\vendor\laravel-shift\blueprint\src\BlueprintCommand.php:47

  2   Blueprint\BlueprintCommand::handle()
      D:\xampp\htdocs\websockets\vendor\laravel\framework\src\Illuminate\Container\BoundMethod.php:32

  Please use the argument -v to see more details.
```

I just want to know where to create the `draft.yaml` inside the Laravel Application

jasonmccreary:
Thanks. I'll get this cleaned up before version 1.0.

jasonmccreary:
The `draft.yaml` file should be the same level of `artisan`. I have added the file check and better error message. This will also be part of the documentation.

PR #21, opened 2019-11-20
Fix draft file not found

ryangjchandler:
This should make #15 a bit more obvious. By default, it will look for the 'draft.yaml' file in the current working directory which if you're running 'php artisan' will be the root of your Laravel project.

You could also use a relative path such as '../draft.yaml' and this would work too I believe. I imagine there will tests for this in the future.

### DIFF
--- a/src/BlueprintCommand.php
+++ b/src/BlueprintCommand.php
@@ -44,7 +44,11 @@ class BlueprintCommand extends Command
      */
     public function handle()
     {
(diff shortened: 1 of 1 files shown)


jasonmccreary:
I'm not sure I understand the need to get the current working directory. Any file path provided should be relative to `artisan`.

However, the check is definitely helpful. I'm going to close this, but `cherry-pick` that commit. Thanks!

PR #34, opened 2019-12-07
Feature/add erase command

### LINKED ISSUE #16: Undo previously built components
While it's easy to use a clean working state and version control to manage the generation of new components, it would be better if Blueprint had an "undo" command to revert/remove the generated components of the last _built_.

Naoray:
closes #16 

---

I've first considered adding an `erase()` method to all generator-classes which un-does what they generated before. I didn't have a solution to undo just the last changes. Also, it seemed tedious and I had another idea.

Why not just store the generated build in a `yaml` file and iterate over it when the `erase` command is called? The only downside of this approach is that I have no way to access what happened to those files who weren't `created` but just `updated`. Maybe you have a solution for this as well?

EDIT:
I'd add tests for the `EraseCommand` if you would allow me to add the `orchestra/testbench`

### DIFF
--- a/src/Blueprint.php
+++ b/src/Blueprint.php
@@ -2,10 +2,9 @@
 
 namespace Blueprint;
 
-
-use Blueprint\Contracts\Generator;
 use Blueprint\Contracts\Lexer;
 use Symfony\Component\Yaml\Yaml;
+use Blueprint\Contracts\Generator;
 
 class Blueprint
 {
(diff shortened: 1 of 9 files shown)


jasonmccreary:
This is awesome. I planned on making a`.blueprint` file for tracking these types of things. So let me review this and likely amend it to get it merged as a start for the `erase` command "undoing" created files.

jasonmccreary:
I am closing this to reopen with your work on a new, non-conflicting branch and making some adjustments.

PR #43, opened 2019-12-20
Add erase command

jasonmccreary:
This will add a basic `blueprint:erase` to remove the generated files from the last `blueprint:build` command.

This can not reliably erase changes made to existing files. So while this command is helpful, running `blueprint:build` from a clean working state and using version control to erase changes will always be best.


jasonmccreary:
Closes #16

PR #44, opened 2019-12-20
Register Blueprint as singleton for extensibility

jasonmccreary:
By registering Blueprint as a singleton, users may completely override their own implementations or resolve the shared instance to register their own generators.


jasonmccreary:
This is built upon #11 and closes #2.

PR #46, opened 2020-01-08
Add support for longtext in fakeableTypes

RyanPriceDotCa:
Noticed longtext type results in an empty faker value in the factory generation.


jasonmccreary:
Thanks!

PR #48, opened 2020-01-11
Fix the migration datatype of an `id` column

nathane:
If applied, this pull request will:
- [x] change the dataType of an `id` column from `increments` to `bigIncrements` ensuring relationships created with `unsignedBigInteger` _(`UNSIGNED BIGINT equivalent column`)_ matches the dataType of an `id` column _(`Auto-incrementing UNSIGNED BIGINT` )_.
- [x] update all the current test fixtures

| Column Types| Description | 
| --: | :-- |
|`increments` | Auto-incrementing UNSIGNED INTEGER (primary key) equivalent column.|
|`bigIncrements` | Auto-incrementing UNSIGNED BIGINT (primary key) equivalent column.|
|`unsignedBigInteger` | UNSIGNED BIGINT equivalent column.|

### DIFF
--- a/src/Generators/MigrationGenerator.php
+++ b/src/Generators/MigrationGenerator.php
@@ -58,7 +58,7 @@ class MigrationGenerator implements Generator
         foreach ($model->columns() as $column) {
             $dataType = $column->dataType();
             if ($column->name() === 'id') {
-                $dataType = 'increments';
+                $dataType = 'bigIncrements';
             } elseif ($column->dataType() === 'id') {
                 $dataType = 'unsignedBigInteger';
             }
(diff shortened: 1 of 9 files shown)


jasonmccreary:
Nice. Can't believe this wasn't noticed earlier.

Ultimately I want to use the new `id` shorthand built-in to Laravel. But I'm going to support 5.8 a bit longer.

msonowal:
@nathane  @jasonmccreary ,
But now there is no option to have a column name id with normal increments as it always takes bigIncrements which is a issue

jasonmccreary:
@msonowal, not sure your goal, but you can create a column named `id` as an `increments` column explicity.

For example:

```yaml
models:
    User:
        id: increments
```

If that doesn't work, please open an issue.

msonowal:
@jasonmccreary  it does not work that is why at first I have created the issue and opened a PR #107

PR #49, opened 2020-01-17
Added support to Lumen 6.2.0

mateo2181:


src/BlueprintCommand.php
```diff
@@ -50,7 +50,7 @@ public function handle()
         }
 
         $contents = $this->files->get($file);
-        $blueprint = resolve(Blueprint::class);
+        $blueprint = $this->resolve(Blueprint::class);
```
jasonmccreary: If this is all that is needed to support Lumen, please simply change the `resolve` helper to the `app` helper and I'll merge it.

mateo2181: Jason, I realized that Lumen not support FormRequest, so the pull is incomplete.

PR #53, opened 2020-01-24
Class PHPDoc IDE helper

itrack:
I added class comments to automatically identify columns for models.


src/Lexers/ModelLexer.php
```diff
@@ -82,6 +82,69 @@ class ModelLexer implements Lexer
...
+        'nullablemorphs' => 'string',
+        'nullableuuidmorphs' => 'string',
+        'nullabletimestamps' => 'Carbon',
+        'point' => 'string',
+        'polygon' => 'string',
+        'remembertoken' => 'string',
+        'set' => 'string',
+        'smallincrements' => 'integer',
+        'smallinteger' => 'integer',
+        'softdeletes' => 'Carbon',
```
jasonmccreary: Carbon should be fully qualified, i.e. `\Carbon\Carbon`.

src/Generators/ModelGenerator.php
```diff
@@ -53,6 +54,18 @@ protected function populateStub(string $stub, Model $model)
         return $stub;
     }
 
+    private function buildClassPhpDoc(Model $model)
+    {
+        $phpDoc = '/**' . PHP_EOL;
+        /** @var Column $column */
+        foreach ($model->columns() as $column) {
+            $phpDoc.= " * @property {$column->phpDocDataType()} \${$column->name()}" . PHP_EOL;
```
jasonmccreary: This would be lighter weight if the DocBlock data type were inferred here from data type and not added to the Lexer/Column as it has no other value within Blueprint other than this generator.

src/Lexers/ModelLexer.php
```diff
@@ -82,6 +82,69 @@ class ModelLexer implements Lexer
         'unique' => 'unique',
     ];
 
+    private static $phpDocDataTypes = [
+        'id' => 'integer',
```
jasonmccreary: This should be `int` not `integer` - see [Types](https://docs.phpdoc.org/latest/guides/types.html)

src/Lexers/ModelLexer.php
```diff
@@ -82,6 +82,69 @@ class ModelLexer implements Lexer
...
+        'bigincrements' => 'integer',
+        'biginteger' => 'integer',
+        'binary' => 'string',
+        'boolean' => 'boolean',
```
jasonmccreary: Similarly, this should be `bool`.

src/Generators/ModelGenerator.php
jasonmccreary: Still would like to see this done. Then I'm glad to merge.

If you don't have time, no worries. I will use the work you have done as a base and finish this up Tuesday.

jasonmccreary:
Going to close this and finish it up myself.

jasonmccreary:
Finished in #57.

PR #70, opened 2020-03-04
fix: mismatched key casing

nathane:
`FactoryGenerator.php` correctly generates `$faker->text` for `longtext`.

### DIFF
--- a/tests/fixtures/definitions/post.bp
+++ b/tests/fixtures/definitions/post.bp
@@ -2,6 +2,7 @@ models:
   Post:
     title: string
     author_id: id
+    author_bio: longtext
     content: bigtext nullable
     published_at: timestamp nullable
     word_count: integer unsigned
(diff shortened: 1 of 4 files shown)


tests/fixtures/definitions/post.bp
```diff
@@ -2,6 +2,7 @@ models:
   Post:
     title: string
     author_id: id
+    author_bio: longtext
```
jasonmccreary: I would have assumed this should be `longText` to emphasize the point. I've forgotten, does the underlying datatype get proper cased in the `Model`?

nathane: inside `ModelLexer::$dataTypes` https://github.com/laravel-shift/blueprint/blob/031b8030605b8cf58dfc0e7940e43ce7f8c6d076/src/Lexers/ModelLexer.php#L32

PR #72, opened 2020-03-14
Generate model relationships

peterjaap:
I took a stab at #4.

You said:

> I want to determine a streamlined syntax for the latter as well as supporting existing models.

I'm not sure what I've done correlates to what you had envisioned, but this might be a push in the right direction.

Here's an example:

```
models:
  Order:
    user_id: id
    subscription_id: id
    price: float
  Subscription:
    user_id: id
    title: string
    description: longtext
    price: float
    relationships: hasMany:order_id,hasOne:duration_id
  Duration:
    title: string
    months: int
```

When the word 'many' appears in the hasMany/hasOne part of the relationships line, the string between the following colon and `_id` will be pluralized.

This generates this Plan model;

```
<?php

namespace App;

use Illuminate\Database\Eloquent\Model;

class Plan extends Model
{
    /**
     * The attributes that are mass assignable.
     *
     * @var array
     */
    protected $fillable = [
        'user_id',
        'title',
        'description',
        'price',
    ];

    /**
     * The attributes that should be cast to native types.
     *
     * @var array
     */
    protected $casts = [
        'id' => 'integer',
        'user_id' => 'integer',
        'price' => 'float',
    ];


    public function user()
    {
        return $this->belongsTo(\App\User::class);
    }

    public function orders()
    {
        return $this->hasMany(\App\Order::class);
    }
    
    public function duration()
    {
        return $this->hasOne(\App\Duration::class);
    }
}
```

I haven't implemented it yet but "through" relations could be done like this;

```
  Subscription:
    ...
    relationships: hasManyThrough:order_id:payment_id
```

I renamed the existing `buildRelationships` to `buildBelongsTo` because the `belongsTo` is a bit more special case. I then created a new `buildRelationships` to build the other relationships (not being belongsTo).

Let me know what you think!

### DIFF
--- a/src/Generators/FactoryGenerator.php
+++ b/src/Generators/FactoryGenerator.php
@@ -63,7 +63,7 @@ class FactoryGenerator implements Generator
 
         /** @var \Blueprint\Models\Column $column */
         foreach ($model->columns() as $column) {
-            if ($column->name() === 'id') {
+            if ($column->name() === 'id' || $column->name() === 'relationships') {
                 continue;
             }
 
--- a/src/Generators/MigrationGenerator.php
+++ b/src/Generators/MigrationGenerator.php
@@ -61,6 +61,8 @@ class MigrationGenerator implements Generator
                 $dataType = 'bigIncrements';
             } elseif ($column->dataType() === 'id') {
                 $dataType = 'unsignedBigInteger';
+            } elseif ($column->name() === 'relationships') {
+                continue;
             }
 
             $definition .= self::INDENT . '$table->' . $dataType . "('{$column->name()}'";
--- a/src/Generators/ModelGenerator.php
+++ b/src/Generators/ModelGenerator.php
@@ -158,7 +158,7 @@ class ModelGenerator implements Generator
                 if ('belongsTo' === $methodName) {
                     throw new \Exception('The belongsTo relationship for the '.$modelName.' model on the '.$model->name().' model should be defined using the '.$modelName.'_id: id syntax');
                 }
-                if (stripos($methodName, 'many') !== false) {
+                if (Str::contains($methodName, 'many')) {
                     $modelName = Str::plural($modelName);
                 }
                 $class = Str::studly($column->attributes()[0] ?? $modelName);
--- a/src/Generators/ModelGenerator.php
+++ b/src/Generators/ModelGenerator.php
@@ -46,6 +46,7 @@ class ModelGenerator implements Generator
 
         $body = $this->buildProperties($model);
         $body .= PHP_EOL . PHP_EOL;
+        $body .= $this->buildBelongsTo($model);
         $body .= $this->buildRelationships($model);
 
         $stub = str_replace('// ...', trim($body), $stub);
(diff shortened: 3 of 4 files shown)


peterjaap:
(btw the failing PHP Composer / build check has to do with this; https://github.com/michaelw90/PHP-Lint/pull/2)

jasonmccreary:
@peterjaap this is pretty cool.

A few questions:

- Does this create the Model relationships for both sides? For example, a [many to many relationship](https://laravel.com/docs/5.7/eloquent-relationships#many-to-many).
- When necessary, does this generate a migration for the pivot table?

Also, what do you think about this syntax?

```yaml
  Subscription:
    user_id: id
    title: string
    description: longtext
    price: float
    relationships: 
        hasMany:Order
        hasOne:Duration
```

While this makes the definition _feel_ more verbose, it's more readable than a long line. I don't think uses will have many relationships anyway.

Also, using the model name directly would provide a better reference and promote conventions.

Let me know your thoughts. Glad to pair up on this.

peterjaap:
@jasonmccreary no it doesn't do both sides (yet), and no pivot, I wanted to settle on a syntax first with you. Plus, selfishly, my itch only included a `hasMany` for now :)

I thought of that syntax too, but ran into an error about the lexer needing a string instead of an array and decided to pursue the string route first (because I only had an hour of time yesterday). But yes, I'm totally for the array-syntax.

Using the model name directly feels more fluent, I agree.

Maybe we can also throw an error when someone tries to include `belongsTo` in the `relationships` part, saying they should use the `model_id: id` notation instead.

peterjaap:
@jasonmccreary ok that was easier than expected.

I now use this;

```
models:
  Order:
    user_id: id
    subscription_id: id
    price: float
  Subscription:
    user_id: id
    title: string
    description: longtext
    price: float
    relationships: 
      hasMany: order_id
      hasOne: duration_id
  Duration:
    title: string
    months: int
```

Output is the same.

jasonmccreary:
Awesome. Give me a bit to review. I'll merge this in later in the week with some other stuff that's been pending.

tectiv3:
In current implementation you can only specify one relationship of each type.
e.g. this won't work:
```yaml
    relationships:
      hasMany:
        - post
        - comment
```
Fixed it by replacing this block:
https://github.com/laravel-shift/blueprint/pull/72/files#diff-f600277501636eb665e338bf8b4d3ab3R157-R171
with:
```php
            foreach ($column->attributes() as $methodName => $value) {
                if ('belongsTo' === $methodName) {
                    throw new \Exception('The belongsTo relationship for the '.$value.' model on the '.$model->name().' model should be defined using the '.$value.'_id: id syntax');
                }

                $models = Arr::wrap($value);
                foreach ($models as $modelName) {
                    $class = Str::studly($column->attributes()[0] ?? $modelName);
                    $relationship = sprintf("\$this->%s(%s::class)", $methodName, '\\' . $model->fullyQualifiedNamespace() . '\\' . $class);

                    $modelNameForMethod = Str::contains($methodName, 'Many') ? Str::plural($modelName) : $modelName;
                    $method = str_replace('DummyName', Str::camel($modelNameForMethod), $template);
                    $method = str_replace('null', $relationship, $method);

                    $methods .= PHP_EOL . $method;
               }
            }
```
And added `use Illuminate\Support\Arr;` to the imports.

jasonmccreary:
@tectiv3 that you for this. Excellent catch. However, at this point I do feel like the nesting is getting a bit deep.

Would you be willing to update your code block to use a comma delimited format instead?

tectiv3:
Sure, I just liked native yaml array parsing but we can just as easily split it ourselves.

jasonmccreary:
@tectiv3 I understand. I trying to balance the syntax being too dense. There are plenty of places where Blueprint already splits lines. So I think it'll be okay. I also don't see there being too many of each relationship.

peterjaap:
I'm in favor of the native YAML notation instead of comma split strings!

jasonmccreary:
It just makes it super long.

```yaml
models:
    Model:
        column: type
        relationships:
             hasmany:
                 - Order
```

In addition, no where else does Blueprint use the list YAML syntax.

Feel free to submit this as is. I'll change it after.

PR #75, opened 2020-03-23
Add ability to generate API resource controllers

tectiv3:
Hi!
I've extracted API Resource related changes from my fork into this PR.

Please take a look, does everything makes sense?

Tried to keep unrelated changes to the minimum, had to add prettier config so my editor wouldn't reformat the whole file on every save :)

PR #76, opened 2020-03-23
Add uuid, json data types and index/foreign modifiers

tectiv3:
Added support for UUID data type for id columns.
Added support for `index` modifier.
Added support for `foreign` modifier - generates corresponding migration.
Added basic json column data type support.


jasonmccreary:
@tectiv3 thanks for this. Do you have any tests for these changes?

jasonmccreary:
@tectiv3 I'm going to go ahead and merge this and backfill the tests myself. However, given the size of #75, I would appreciate if you could take a stab at writing some tests for it.

PR #84, opened 2020-03-27
Factory generator create folder if missing

justinwhite15:


jasonmccreary:
Looks like this cases a lot of tests to fail.

Also, whats the use case of a nested model factory?

justinwhite15:
When created a nested model in the YAML, Blueprint will attempt to create a factory that is also nested in the same directory/path, in the same way that it is doing for the models/controllers/tests/etc...  But, it can fail because it isn't checking for the existence of the folder first (like it is in the ControllerGenerator).

Re: the tests...I am running into some issues running the tests locally (i.e., I can't get the tests to pass locally with unmodified code).  Once I figure that out I am hoping to track down what is going on with the tests.

justinwhite15:
To reproduce, you can use below as the blueprint yaml file.

```yaml
models:
  Book\Phone:
    label: string
    user_id: id
    phone_number: text
    type: enum:home,cell
    status: set:archived,deleted
```

When running `blueprint:build` it will throw a an ErrorException 
file_put_contents(database/factories/Book/PhoneFactory.php): failed to open stream: No such file or directory

justinwhite15:
I figured out my locally testing issues, so I can now reproduce the same errors thrown via Github Action `vendor/bin/phpunit`

justinwhite15:
I got the tests updated; I hadn't updated the mocks to accompany the change I made to the FactoryGenerator.

justinwhite15:
This is ready for review now; tests pass and I validated that using the yaml below now generates models/factories without error.

```yaml
models:
  Book\Phone:
    label: string
    user_id: id
    phone_number: text
    type: enum:home,cell
    status: set:archived,deleted
```

Result
```sh
Created:
- database/migrations/2020_03_27_213306_create_phones_table.php
- app/Book/Phone.php
- database/factories/Book/PhoneFactory.php
```

If this isn't the desired behavior, I'd be happy to look into changing the behavior of the FactoryGenerator.

justinwhite15:
One additional thing I noticed, the `FactoryGeneratorTest.php` uses the modelTreeDataProvider, and one of the elements is
```php
            ['definitions/nested-components.bp', 'database/factories/Admin/UserFactory.php', 'factories/nested-components.php'],
```

This was expecting the factory to be generated in a nested folder.  If I am understanding what it was doing properly, it appears as though the tests pass because the mock file system said the file existed, but wasn't checking/accounting for the folder not existing first.

jasonmccreary:
Yeah. Something is strange here. I'd like to add a new tests case (maybe the `Book` example from your comments) as well as ensuring the `definitions/nested-components.bp` one works too.

I know you have the right code. But I want to ensure it's driven out by the tests so there's coverage for the future.

Give it one more pass if you can, otherwise, I'll clean it up over the weekend.

jasonmccreary:
@justinwhite15 thanks again for this. I went ahead and merged and backfilled an explicit test [commit: 49444de](https://github.com/laravel-shift/blueprint/commit/49444de64b713f3c530a3d84ae8ed226b43062e6)

PR #97, opened 2020-04-13
Include password field in fillable model properties

### LINKED ISSUE #92: Column called password does not get added to fillable array
After creating the basic user model, i have noticed that Password does not get added, in a default user model it gets created as -

```
    protected $fillable = [
        'name', 'email', 'password'
    ];
```

However using blueprint, we get the following -

```
    protected $fillable = [
        'name',
        'email',
        'is_active',
        'email_verified_at',
        'remember_token',
    ];
```

It also totally misses the hidden block in a normal user model -

```
    protected $hidden = [
        'password', 'remember_token'
    ];
```

adevade:
It's fillable with Laravel's default User.php model.

Closes #92

### DIFF
--- a/src/Generators/ModelGenerator.php
+++ b/src/Generators/ModelGenerator.php
@@ -151,7 +151,6 @@ class ModelGenerator implements Generator
     {
         return array_diff(array_keys($columns), [
             'id',
-            'password',
             'deleted_at',
             'created_at',
             'updated_at',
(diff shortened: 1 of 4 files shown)


jasonmccreary:
Could you update or create a model fixture to include this so it's covered by a test.

jasonmccreary:
You'll need to rebase this now that #98 was merged.

adevade:
I'm not much of a git wizard... Is this good enough?

PR #98, opened 2020-04-13
Add $hidden attribute to Models

adevade:
This PR will add a `$hidden` attribute to models if the columns include `password` or `remember_token`. The same values as in the default Laravel `User.php`.

Mentioned in #92

### DIFF
--- a/src/Generators/ModelGenerator.php
+++ b/src/Generators/ModelGenerator.php
@@ -99,6 +99,11 @@ class ModelGenerator implements Generator
             $properties .= $this->files->stub('model/fillable.stub');
         }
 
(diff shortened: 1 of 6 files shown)


src/Generators/ModelGenerator.php
```diff
@@ -158,6 +163,16 @@ private function fillableColumns(array $columns)
         ]);
     }
 
+    private function hiddenColumns(array $columns)
+    {
+        return array_filter(array_keys($columns), function ($column) {
```
jasonmccreary, line 168: Could use `array_intersect`.

adevade: Done! ⚡

PR #99, opened 2020-04-13
[WIP] Generate id() column on Laravel 7 and up

### LINKED ISSUE #79: Support `id` column data type for Laravel 7
This adds support of the `id` shorthand column type added in Laravel 7.

Since Blueprint will continue to support back to Laravel 6 (LTS) for a while, this should only do so for apps running Laravel 7 or higher.

adevade:
No tests for this one yet, and maybe not a good looking implementation, but the output is right!

Closes #79 if merged. Feel free to continue work.

### DIFF
--- a/src/Generators/MigrationGenerator.php
+++ b/src/Generators/MigrationGenerator.php
@@ -62,7 +62,11 @@ class MigrationGenerator implements Generator
                 $dataType = 'uuid';
             }
 
-            $definition .= self::INDENT . '$table->' . $dataType . "('{$column->name()}'";
+            if ($this->isLaravel7orNewer() && $dataType === 'bigIncrements') {
+                $definition .= self::INDENT . '$table->id(';
+            } else {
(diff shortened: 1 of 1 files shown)


jasonmccreary:
I'll have to figure out a way to toggle the app version. I'm sure there's probably a way to do it with TestBench. But don't know off the top of my head.

src/Generators/MigrationGenerator.php
```diff
@@ -117,4 +121,9 @@ protected function getPath(Model $model, Carbon $timestamp)
     {
         return 'database/migrations/' . $timestamp->format('Y_m_d_His') . '_create_' . $model->tableName() . '_table.php';
     }
+
+    protected function isLaravel7orNewer()
+    {
+        return version_compare(app()->version(), '7.0.0', '>=');
```
nathane, line 127: current laravel version via TestBench `$this->app::VERSION`

jasonmccreary: Yeah, I was trying to see if there was a way to fake it.

PR #103, opened 2020-04-14
support space between quotes in `default` modifier

### LINKED ISSUE #91: Default text cant have a space between words
```
  BillingProfile:
    invoice_country: string default:"United Kingdom"
```

With the space inbetween the two, it gets to "United but then skips as it sees a space.

```
        Schema::create('billing_profiles', function (Blueprint $table) {
            $table->string('invoice_country')->default("United);
        });
```

I have tried this with both Single and Double quotes to the same effect.

nathane:
close #91

### DIFF
--- a/src/Lexers/ModelLexer.php
+++ b/src/Lexers/ModelLexer.php
@@ -181,7 +181,7 @@ class ModelLexer implements Lexer
         $data_type = 'string';
         $modifiers = [];
 
-        $tokens = explode(' ', $definition);
+        $tokens = preg_split('#".*?"(*SKIP)(*F)|\s+#', $definition);
         foreach ($tokens as $token) {
             $parts = explode(':', $token);
             $value = $parts[0];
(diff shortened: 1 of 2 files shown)


jasonmccreary:
I suppose all `default` values with a space would be quoted, but wonder if there are other formats we're not thinking of that would cause another issue. 

Maybe sending this _modifier_ through the YAML parser would be the best bet...

nathane:
These are the different formats i was able to find. 
[source](https://github.com/search?l=PHP&q=%22Blueprint+%24table%22+AND+%22default%22&type=Code)
```
->default(\DB::raw('CURRENT_TIMESTAMP'));
->default(\DB::raw('CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP'));
->default(null);
->default(0);
->default(1);
->default(true);
->default(false);
->default('text');
->default(0.00); // decimal
->default('0000-00-00'); // date
``` 
and 


> The default modifier accepts a value or an \Illuminate\Database\Query\Expression instance. Using an Expression instance will prevent wrapping the value in quotes and allow you to use database specific functions. One situation where this is particularly useful is when you need to assign default values to JSON columns: https://laravel.com/docs/7.x/migrations
```
$table->json('movies')->default(new Expression('(JSON_ARRAY())'));
```
> Support for default expressions depends on your database driver, database version, and the field type. Please refer to the appropriate documentation for compatibility. Also note that using database specific functions may tightly couple you to a specific driver.

jasonmccreary:
I think Blueprint is "good enough" to support simple values and strings. Maybe a shorthand of `now` or something. But not the `DB::raw` other expressions. So this should be enough for now.

PR #104, opened 2020-04-14
support belongsToMany relationship

raftalks:
This PR introduces belongsToMany relationship. 
Example usage syntax:
```
Team:
    name: string
    description: longtext
    relationships: 
        belongsToMany:User

User:
    username: string
    relationships: 
        belongsToMany:Team
```

Generates migration pivot table and adds relationship to the model


jasonmccreary:
Thanks for this. I'll review this evening and add some tests as I will need to do so before accepting.

jasonmccreary:
Thank you for the start. Looks like another contributor has used this and added the missing tests. So I have merged your work through #113.

PR #105, opened 2020-04-15
Add PHPDoc for model relationships

amir9480:
Hello.

I added PHPDoc with return type to model relationships.
It can help code editor provide autocomplete.

### DIFF
--- a/src/Generators/ModelGenerator.php
+++ b/src/Generators/ModelGenerator.php
@@ -121,6 +121,7 @@ class ModelGenerator implements Generator
     {
         $methods = '';
         $template = $this->files->stub('model/method.stub');
+        $commentTemplate = $this->files->stub('model/method-comment.stub');
 
         foreach ($model->relationships() as $type => $references) {
             foreach ($references as $reference) {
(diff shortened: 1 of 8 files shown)


jasonmccreary:
@amir9480, why did you close this? Seems like a good addition. However, I would like to make it conditional based on the `generate_phpdocs` config option...

amir9480:
@jasonmccreary Sorry.
I struggling with unit tests on my machine and can not run PHPUnit tests locally because of some errors like this.
```
Mockery\Exception\NoMatchingExpectationException: No matching handler found for Mockery_4::put
```

I want to make sure unit tests pass first.

amir9480:
@jasonmccreary If you have any similar experience with the exception or have any clue why this happening I appreciate it if you let me know.
I still didn't find why this happens.

jasonmccreary:
@amir9480, you'll likely need to `expect` the call you added to get the new stub. Otherwise, the mocked `file` property will throw this exception.

amir9480:
For anyone who comes from the future.

If you using Windows and also docker environment for tests make sure your git config `core.autocrlf` is false. (Mine was true).
Also, make sure your IDE doesn't convert LF to CRLF automatically.
This takes about 4 hours for me to find out 😅.

PR #106, opened 2020-04-15
Add PHPDoc for model relationships

amir9480:
Hello.

I added PHPDoc with return type to model relationships.
It can help code editor provide autocomplete.


jasonmccreary:
I think this should be good. However, you'll need to rebase and resolve the conflicts.

Also, can you have this honor the `blueprint.generate_phpdocs` config option as the `buildClassPhpDoc` does. That way these are only output when this option is `true` and not by default.

amir9480:
@jasonmccreary 

Hello,
I added `blueprint.generate_phpdocs` config check and also added a file for unit tests in `tests/fixtures/models/relationships-phpdoc.php`.

https://github.com/amir9480/blueprint/blob/9120aa353b575a6e53770c75621760fc4e051a90/tests/Feature/Generator/ModelGeneratorTest.php#L226

https://github.com/amir9480/blueprint/blob/add_model_relation_phpdoc/tests/fixtures/models/relationships-phpdoc.php

PR #107, opened 2020-04-15
Added support for unsigned integer(increments) column type

msonowal:
Added support for unsigned integer column types with increments which was removed by the PR
https://github.com/laravel-shift/blueprint/pull/48

### DIFF
--- a/src/Generators/MigrationGenerator.php
+++ b/src/Generators/MigrationGenerator.php
@@ -55,7 +55,7 @@ class MigrationGenerator implements Generator
         /** @var \Blueprint\Models\Column $column */
         foreach ($model->columns() as $column) {
             $dataType = $column->dataType();
-            if ($column->name() === 'id' && $dataType !== 'uuid') {
+            if ($column->name() === 'id' && $dataType !== 'uuid' && $dataType !== 'increments') {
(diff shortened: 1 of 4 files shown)


jasonmccreary:
I see the bug here. But this solution is not universal.

I think the correct logic is only when the name is `id` and the type is `id`.

msonowal:
@jasonmccreary 
At first I thought like that but seems like as there are conditions are added for more specific case so I thought it will be a global fix for only the id column

jasonmccreary:
If you change it like I noted, do all the tests still pass? Where is the test for your use case?

msonowal:
Hi @jasonmccreary 
I did tested in local it does pass everything, But I am not sure for writing the test specific for checking to column types as there no example present(I may be wrong), if you can enlighten me with some reference it will be very much helpful

jasonmccreary:
You should be able to look at the `MigrationGeneratorTest` and add a fixture with your test case.

nathane:
@msonowal sorry about the bug I introduced in #48, this #109 should fix it.


cc: @jasonmccreary

msonowal:
cool @nathane ,

But I was already preparing the tests anyways thanks

PR #109, opened 2020-04-15
fix: datatype of an `id` column

nathane:
fix the bug I created in #48

### DIFF
--- a/src/Generators/MigrationGenerator.php
+++ b/src/Generators/MigrationGenerator.php
@@ -57,7 +57,7 @@ class MigrationGenerator implements Generator
             $dataType = $column->dataType();
             if ($column->name() === 'id' && $dataType === 'id') {
                 $dataType = 'bigIncrements';
-            }elseif ($column->name() !== 'id' && $dataType === 'id') {
+            } elseif ($dataType === 'id') {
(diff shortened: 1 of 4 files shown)


src/Generators/MigrationGenerator.php
```diff
@@ -55,12 +55,10 @@ protected function buildDefinition(Model $model)
         /** @var \Blueprint\Models\Column $column */
         foreach ($model->columns() as $column) {
             $dataType = $column->dataType();
-            if ($column->name() === 'id' && $dataType !== 'uuid') {
+            if ($column->name() === 'id' && $dataType === 'id') {
                 $dataType = 'bigIncrements';
-            } elseif ($column->dataType() === 'id') {
+            }elseif ($column->name() !== 'id' && $dataType === 'id') {
```
jasonmccreary, line 60: This is unnecessary given the `elseif`

nathane: it's needed to build relationships. `$column->name() = 'user_id'` and `$dataType === 'id'`

jasonmccreary:
Thanks all!

msonowal:
@jasonmccreary 
Can you tag a release

PR #110, opened 2020-04-15
Optimize unsigned columns

### LINKED ISSUE #5: Optimize column datatypes
When defining models it would be good to _optimize_ any column data types and modifiers.

For example, the definition `column: integer unsigned` currently generates `$table->integer('column')->unsigned()`, but could be optimized to generate `$table->unsignedInteger('column')`.

Additional optimizations can be inferred from the [list of available data types](https://laravel.com/docs/5.8/migrations#creating-columns).

willvincent:
Any unsignable datatype that has the unsigned modifier specified will be optimized to it's `unsigned[type]` corresponding datatype.

ie: 
```
foo: integer unsigned
```

Becomes:
```
$table->unsignedInteger('foo');
```

And so forth.

Closes #5

### DIFF
--- a/src/Generators/MigrationGenerator.php
+++ b/src/Generators/MigrationGenerator.php
@@ -12,6 +12,15 @@ class MigrationGenerator implements Generator
 {
     const INDENT = '            ';
 
+    const UNSIGNABLE_TYPES = [
+      'bigInteger',
+      'decimal',
+      'integer',
+      'mediumInteger',
+      'smallInteger',
+      'tinyInteger',
+    ];
+
     /** @var \Illuminate\Contracts\Filesystem\Filesystem */
     private $files;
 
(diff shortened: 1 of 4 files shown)


jasonmccreary:
Very nice. Are there any others you think, beyond `unsigned`?

willvincent:
Guess there are a couple nullable too.. nothing else jumps out at me. Let me get those added though

willvincent:
Omitted `nullableTimestamps()` as that's the same as `timestamps()` and arguable less optimal :)

willvincent:
Good call.. `Str::StartsWith` is more readable

jasonmccreary:
Yeah. I'll likely optimize the `continue` as well, but this is a great start.

Issue #85, opened 2020-03-27
id: false doesn't disable the id column


justinwhite15:
Documentation states:

> By default, each model will automatically be defined with an id and timestamps columns. To disable these columns you may define them with a false value. For example, timestamps: false.

But adding `id: false` to a model's definition in the YAML doesn't disable the creation of the id column in the migration.

To reproduce, use the .yaml below:
```yaml
models:
  Phone:
    id: false
    label: string
    user_id: id
    phone_number: text
    type: enum:home,cell
    status: set:archived,deleted
    timestamps: false
```

This will generate the migration below:
```php
        Schema::create('phones', function (Blueprint $table) {
            $table->bigIncrements('id');
            $table->string('label');
            $table->unsignedBigInteger('user_id');
            $table->text('phone_number');
            $table->enum('type', ["home","cell"]);
            $table->set('status', ["archived","deleted"]);
        });
```

I took a quick look through the code, and although the Model model does include a usesTimestamps() method, it doesn't have a usesId() method.

Options
1. I am using the wrong YAML
2. Update the documentation to indicate that the id field cannot be disabled
3. Update the code to allow for an id field to be not generated.

@jasonmccreary Do you have a preference? Or, did I miss something?

jasonmccreary:
@justinwhite15, totally agree, `id: false` should be a thing.

I actually ran into this when recording my Laracasts demo and didn't open an issue for it. 😅

ttomdewit:
Working on it as we speak.

ttomdewit:
I quickly scanned the existing code to disable timestamps and tried to mimic it's behaviour in https://github.com/laravel-shift/blueprint/pull/111.

I need to add a migration test that verifies that ID columns are indeed ignored.

PR #111, opened 2020-04-15
Initial attempt to disable IDs

ttomdewit:
Initial attempt to disable IDs which closes https://github.com/laravel-shift/blueprint/issues/85


ttomdewit:
- [ ] Add test that verifies ID column is indeed not created

tests/fixtures/definitions/without-id.bp
```diff
@@ -0,0 +1,8 @@
+models:
+  Name:
+    softDeletes
```
jasonmccreary, line 3: How does this test `id: false`?

ttomdewit: Ah yes. I copied the original test and stub over to the new test and needed some attributes to remain. Figured I’d only delete the ID attribute.

jasonmccreary: I guess I don't understand what this test proves. I thought you were trying to support `id: false` disabling the id field. As such, I'd expect this definition to have `id: false` or the test to fail...

tests/Feature/Lexers/ModelLexerTest.php
```diff
@@ -4,6 +4,7 @@
 
 use Blueprint\Lexers\ModelLexer;
 use Tests\TestCase;
+use function var_dump;
```
jasonmccreary, line 7: Please remove unused function imports.

src/Generators/MigrationGenerator.php
```diff
@@ -69,9 +69,9 @@ protected function buildDefinition(Model $model)
         /** @var \Blueprint\Models\Column $column */
         foreach ($model->columns() as $column) {
             $dataType = $column->dataType();
-            if ($column->name() === 'id' && $dataType === 'id') {
+            if ($column->name() === 'id' && $dataType === 'id' && $model->usesPrimaryKey()) {
                 $dataType = 'bigIncrements';
-            } elseif ($dataType === 'id') {
+            } elseif ($dataType === 'id' && $model->usesPrimaryKey()) {
```
jasonmccreary, line 74: Why does this line need the check? Shouldn't it only be for disabling the `id` column?

tests/fixtures/definitions/without-id.bp
nathane: Omitting `id` should not change the current behavior. Explicitly setting`id: false` should.

jasonmccreary: Exactly. That's why I expect this fixture to have `id: false` based on the test.

ttomdewit: So sorry. I’ll fix this!

ttomdewit: I explicitly set `id: false` and updated the test to prove `'id' => false` is set.

src/Generators/MigrationGenerator.php
ttomdewit: I misunderstood this line in the code.

ttomdewit:
Thank you

jasonmccreary:
Sure. I did [some refactoring](https://github.com/laravel-shift/blueprint/commit/0ff38dd72ef9ca8b3e6cc3f423b729a7f691a78c) after merging if you want to take a look.

ttomdewit:
> Sure. I did [some refactoring](https://github.com/laravel-shift/blueprint/commit/0ff38dd72ef9ca8b3e6cc3f423b729a7f691a78c) after merging if you want to take a look.

I saw that. Will use that when I attempt to add UUID as shorthand later tonight (no promises, but I'm going to try).

PR #117, opened 2020-04-16
feat: generate resource `toArray` for non-hidden columns

### LINKED ISSUE #94: Generate Resource data based on Model
Currently, only the class "stub" is generate for a Resource.

Given that the `resource` statement referenced a model, it would be nice to generate the return value of the `toArray` for all of the _non-hidden_ columns of the model.

nathane:
Close #94


jasonmccreary:
Thanks for taking the extra time to clean this up. Been swamped with PRs this week, so having a smaller set of files to review makes it easier.

I think I'm going to refactor `modelForContext` in tomorrow's live stream. It's not used in multiple generators and needed in more for some of the open issues.

PR #118, opened 2020-04-16
allow custom stub path

Naoray:
This little change helps a lot when writing addons for Blueprint

### DIFF
--- a/config/blueprint.php
+++ b/config/blueprint.php
@@ -56,4 +56,5 @@ return [
     */
     'generate_phpdocs' => false,
 
+    'fake_nullables' => true,
 ];
--- a/src/Generators/FactoryGenerator.php
+++ b/src/Generators/FactoryGenerator.php
@@ -2,9 +2,10 @@
 
 namespace Blueprint\Generators;
 
-use Blueprint\Contracts\Generator;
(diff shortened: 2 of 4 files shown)


jasonmccreary:
I think I get it. But could you explain a little more of _how_ it helps.

Always helps to provide example info in your PRs ;)

Naoray:
Sorry, should have provided an example 🙈 .

My folder structure:
- src/...
- stubs/...

With this PR it would be possible to access all stubs within my packages “stubs“ folder like this:

```php
​$stub​ ​=​ ​$this​->​files​->​stub(​'​class.stub​'​, ​$this​->​stubPath());​
```​

Sure, I could simply access the stub manually but using, the small mixin you created feels much cleaner

Naoray:
In general this change would allow any addon package to utilize the
`->stub()` method to access their own stubs with the possibility to reuse
existing stubs from blueprint

jasonmccreary:
Thanks for the clarification.

I'm going to sit on this for a few days. Someone reported that because this is a mixin/macro it is available from within your App (depending on how you've install the package).

Either way, it feels unclean. As such I may move this to the `Blueprint` namespace.

PR #120, opened 2020-04-16
Update tests namespace

### LINKED ISSUE #119: Tests Namespace Collision
First of all thank you for this!
I'm going to use it for sure and contribute to it.

1 Issue and 1 Suggestion:

- Issue:
Laravel's TestCase class has the same namespace as this package TestCase.

image

- Suggestion:
Add a model schema table with all the options/modifiers so its easy/quick to find all the available options.

nathane:
change `Tests` namespace to `Blueprint\Tests`.

close #119

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Tests\\": "tests/"
+            "Blueprint\\Tests\\": "tests/"
         }
     },
     "extra": {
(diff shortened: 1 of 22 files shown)


jasonmccreary:
Again, thanks for this, but packages it seems odd that the `dev` autoloading is an issue. I think the proper solution would be to _exclude_ these files from export.

Going to a fresh PR with an alternative solution. Will close this for now, but may eventually merge it.

PR #127, opened 2020-04-17
add fake_nullable config

### LINKED ISSUE #78: Exclude `nullable` columns from model factories
In retrospect filling fake data for `nullable` columns seems to be too explicit. Often these columns should be left undefined in the default factory and instead defined in a [factory state](https://laravel.com/docs/5.8/database-testing#factory-states).

This is open for discussion as it would be a _breaking change_ based on original behavior.

Naoray:
closes #78

### DIFF
--- a/config/blueprint.php
+++ b/config/blueprint.php
@@ -56,4 +56,5 @@ return [
     */
     'generate_phpdocs' => false,
 
+    'fake_nullables' => true,
 ];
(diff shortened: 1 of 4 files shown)


Naoray:
sorry for the formatting stuff that changed. Hope you don't mind. Maybe we can add a syleci integration or something similar?

jasonmccreary:
It is added. Just [not automated as part of the PR](https://twitter.com/gonedark/status/1250150048765284357) for whatever reason.

PR #129, opened 2020-04-17
feat: support nullable columns in model factories

nathane:
Use Faker's `optional()` modifier for `nullable` columns.

#78

### DIFF
--- a/src/Generators/FactoryGenerator.php
+++ b/src/Generators/FactoryGenerator.php
@@ -109,6 +109,14 @@ class FactoryGenerator implements Generator
                 $definition .= '$faker->' . $faker;
                 $definition .= ',' . PHP_EOL;
             }
+
+            if (
+                !Str::contains($definition,'optional') &&
+                !in_array($column->dataType(),['id', 'uuid'],true) &&
+                in_array('nullable',$column->modifiers(),true)
+            ){
+                $definition = str_replace('$faker->','$faker->optional()->',$definition);
+            }
         }
 
         return trim($definition);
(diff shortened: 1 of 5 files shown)


jasonmccreary:
I like how this communicates its an _optional` (aka nullable) piece of data. But since this doesn't ensure the data will always be `null`, I'm not sure if it's worth merging...

nathane:
> But since this doesn't ensure the data will always be`null`, 

```
// optional() accepts a weight argument to specify the probability of receiving the default value.
// 0 will always return the default value; 1 will always return the provider. Default weight is 0.5 (50% chance).
$faker->optional($weight = 0.1)->randomDigit; // 90% chance of NULL
$faker->optional($weight = 0.9)->randomDigit; // 10% chance of NULL

// optional() accepts a default argument to specify the default value to return.
// Defaults to NULL.
$faker->optional($weight = 0.5, $default = false)->randomDigit; // 50% chance of FALSE
$faker->optional($weight = 0.9, $default = 'abc')->word; // 10% chance of 'abc'](url)

```

jasonmccreary:
I see. I read the docs too quickly. 

I like how this communicates the `null` value, but quickly allows a developer to remove the `optional` prefix if they want to fake the data.

I'm not sure if this would be considered a "breaking change". It takes a bit of a different meaning with Blueprint as this generates code. Meaning the code it already generated would not be affected. However, it is different behavior moving forward...

Going to wait a bit on this...

jasonmccreary:
Still can't decide on this, so I think it means "close it for now".

While I may not always agree with it, I think faking nullable columns in the default factory definition is the most _helpful_ approach.

PR #131, opened 2020-04-17
fix needles,change slash to backslash

Algeneralo:
Needles should have a backslash, not a slash.

### DIFF
--- a/src/Generators/ControllerGenerator.php
+++ b/src/Generators/ControllerGenerator.php
@@ -221,7 +221,7 @@ class ControllerGenerator implements Generator
         }
 
         $matches = array_filter(array_keys($this->models), function ($key) use ($context) {
-            return Str::endsWith($key, '/'.Str::studly($context));
+            return Str::endsWith($key, '\\'.Str::studly($context));
         });
 
         if (count($matches) === 1) {


jasonmccreary:
I reverted this earlier as I didn't understand the change. Can you demonstrate _why_ or, better yet, write a test.

Algeneralo:
@jasonmccreary An example of why using a backslash:
draft.yaml:
`Models\Post`
So, registered models in `ControllerGenerator` will have a backslash.

Now, you are comparing like this: 
`Str::endsWith("Models\Post", '/Post' )`
This line will never be executed since you're using a slash 
So,it should be `Str::endsWith("Models\Post", '\Post' )`

jasonmccreary:
I see. So the issue here is really the definition of using a forward slash or a back slash.

I think what I'd rather do is normalize that in the Lexer rather than all the generators.

nathane:
have you tried using `Models/Post` instead of `Models\Post` 

https://github.com/laravel-shift/blueprint/blob/0c056fd0223f402c25e8e01c52109db4fddc1e0f/tests/fixtures/definitions/nested-components.bp#L1-L10

Algeneralo:
@nathane yes I did, and it throws an error 
image

jasonmccreary:
Yes. I agree this a bug. But I'm going to patch it when I do the refactor of centralizing the "model lookup".

nathane:
I see. I believe jmac mentioned having to refactor `modelForContext`. It's reused in a few generators.

> I think I'm going to refactor `modelForContext`.
_Originally posted by @jasonmccreary in https://github.com/laravel-shift/blueprint/pull/117#issuecomment-614839120_

jasonmccreary:
@Algeneralo, we should have tests for that. Everything is currently passing. Can you ensure you pulled the last version and try both slashes again?

Algeneralo:
@jasonmccreary Backslash is working correctly now,
but for slash, it's not, still have wrong namespace and throws an error as I mentioned before

jasonmccreary:
Good. I felt we has a test for backslash. In fairness, that is the correct slash to use in the case of namespace.

Nonetheless, I will normalize this during the aforementioned refactor to allow for forward slashes in the definitions.

PR #135, opened 2020-04-19
Fix factory generation for morphs column type

diogogomeswww:
Right now if we use a morphs column type like:
```
foo: morphs
```

We get a factory like this:
image

This PR will add those 2 fields (id and type) for morphs.
For a column definition like `morphs nullable` it will not generate factory fields for those 2 columns.

An `isNullable()` helper method for the Column Model.

Questions:

- Should we add a factory definition for a nullable column? 
In my opinion we shouldn't. If a column is nullable, the factory should not have a definition for that column, otherwise we don't have a way to set it to null. If we send a null value for a key, the factory will generate the value for that key.

- Should we add support for  `nullableMorphs`, `nullableUuidMorphs`, `nullableTimestamps` column definitions that are availabe on Laravel.

So right now this works:
```
foo: morphs nullable
```
But this doesn't
```
foo: nullableMorphs
```

### DIFF
--- a/src/Generators/FactoryGenerator.php
+++ b/src/Generators/FactoryGenerator.php
@@ -107,6 +107,9 @@ class FactoryGenerator implements Generator
                 $default = $column->defaultValue() ?? "'{}'";
                 $definition .= self::INDENT . "'{$column->name()}' => {$default}," . PHP_EOL;
             } elseif ($column->dataType() === 'morphs') {
+                if ($column->isNullable()) {
+                    continue;
+                }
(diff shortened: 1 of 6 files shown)


jasonmccreary:
Thanks for this. Two quick things before merging...

1. Can you rename `nullable` column to `bar` in your test fixture. Was just a bit confusion.
2. Now that you have added the `isNullable` method, can you refactor other areas of the code to use this as well.

diogogomeswww:
Requested changes done.

Issue #136, opened 2020-04-20
Few issues when using belongsToMany


coldKingdom:
In my draft I have setup an relationship like this:
```yaml
relationships:
      belongstoMany: Diary
```

There are a few things that I noticed that makes it unusable

**In the migration: 2020_04_20_111410_create_diary_journey_table.php**:

The class getting the name CreateDiaryJourneyPivotTable.
Pivot probably should not be there or vice versa?

The schema getting things a bit wrong
```php
        Schema::create('diary_journey', function (Blueprint $table) {
            $table->bigIncrements('Journey_id');
            $table->bigIncrements('diary_id');
        });
```
The correct one
```php
        Schema::create('diary_journey', function (Blueprint $table) {
            $table->bigInteger('journey_id');
            $table->bigInteger('diary_id');
        });
```

* bigIncrements should be bigInteger
* Journey_id should be journey_id

Otherwise it works very well as far I've tested it.
Thank you very much for this contribution!

jasonmccreary:
Oh noes. I'll take a look later today to see if we can get this resolved. It was a recent addition and seems to be setup wrong.

Should be some simple changes, if you want to take a crack at the PR. 👍

mikevandiepen:
I don't see "relations" anywhere in the documentation, Is this right?

jasonmccreary:
Yes. You may define `relationships` on models. About 11 stars shy of working on documentation ;)

In the meantime, reference [this comment](https://github.com/laravel-shift/blueprint/pull/72#issuecomment-599174091) for more detail.

PR #139, opened 2020-04-22
Fixes pivot table filename and column references

judereid:
This is a fix for #136 

Get lowercase $segment for column name
Sets the default dataType as bigInteger
Sets the filename to _pivot_table.php - I prefer this convention instead of removing from classname but could go either way there.

### DIFF
--- a/src/Generators/MigrationGenerator.php
+++ b/src/Generators/MigrationGenerator.php
@@ -163,12 +163,12 @@ class MigrationGenerator implements Generator
         return trim($definition);
     }
 
-    protected function buildPivotTableDefinition(array $segments, $dataType = 'bigIncrements')
+    protected function buildPivotTableDefinition(array $segments, $dataType = 'bigInteger')
     {
         $definition = '';
 
         foreach ($segments as $segment) {
-            $column = $segment . '_id';
+            $column = strtolower($segment) . '_id';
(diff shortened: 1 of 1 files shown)


jasonmccreary:
Two things:

- The datatype should be `bigUnsignedInteger` to match other `id` column definitions.
- The migration filename should be the conventional class name of the pivot table. No need for the `_pivot` suffix.

coldKingdom:
> 
> 
> Two things:
> 
>     * The datatype should be `unsignedBigInteger ` to match the `id` column definition.
> 
>     * The migration filename should be the conventional class name of the pivot table (the alphabetical order of the related, singular model names). No need for the `_pivot` suffix.

One thing, are you sure about the second thing? I can't make it work if the file name doesn't match the class. As the class gets the name with pivot in it right now, the migration file also should have pivot in it, right? Like the PR suggests.

Sorry if my English doesn't make sense, it's not my native language :)

jasonmccreary:
@coldKingdom, all good.

Yes, please remove the word "Pivot" from any filename or class name. The conventional table name should be unique enough.

judereid:
@jasonmccreary Cool, I've made those updates now. Yeah, I wasn't sure what the preferred convention for the migration filename was but it makes sense just to match the table name directly - Cheers!

Issue #137, opened 2020-04-21
Use foreignId for belongsTo relationships in Laravel 7


jasonmccreary:
If the user is running Laravel 7, it'd be nice to generate the migration with the proper `foreign` key constraints.

Take a look at [this PR](https://github.com/laravel/framework/pull/32449) for the latest goodness in Laravel 7.

Also, as not all users may want to make foreign key constraints, we should probably introduce a `use_constraints` config option.

mikevandiepen:
I would love to see this, just discovered the project and I'm already loving it!

PR #141, opened 2020-04-24
Worked on #137 (Foreign key constraints for pivot tables)

mikevandiepen:
This PR addresses [#137](https://github.com/laravel-shift/blueprint/issues/137)
**Note, forgot to update the default foreign keys for the belongsTo relation, will fix tomorrow!**

I created foreign key constraints for the pivot tables. I also wrote tests but they failed to execute, didn't have time today to figure out why. There is also an issue where I test whether the Laravel version is `7.0.0` or higher but I failed to execute the test in the testcase since it validates whether `App::version()` is greater or equal to `7.0.0` and I couldn't find a clean way to change this in the test.

I hope you can take a look at this, I made it in a couple of hour and will attempt to look at it again tomorrow, If you think this is sufficient or you can fix the trailing issues yourself you're welcome.


src/Generators/MigrationGenerator.php
```diff
@@ -57,7 +57,7 @@ public function output(array $tree): array
                         continue;
                     }
 
-                    $path = $this->getPivotTablePath($pivotTable, $sequential_timestamp);
+                    $path = $this->getPivotTablePath($pivotTable, $sequential_timestamp->addSecond());
```
jasonmccreary, line 60: This should not be removed. It's likely why the test is failing.

When using constraints, I see why the pivot table would need to be created _after_ the other tables. However, there's really no way to guarantee the order. The best option would be to create all pivot tables _last_.

src/Generators/MigrationGenerator.php
```diff
@@ -156,13 +156,26 @@ protected function buildDefinition(Model $model)
         return trim($definition);
     }
 
-    protected function buildPivotTableDefinition(array $segments, $dataType = 'unsignedBigInteger')
+    protected function buildPivotTableDefinition(array $segments, $dataType = 'foreignId')
```
jasonmccreary, line 159: Is the `dataType` parameter ever sent as something? Just wondering why it's even needed or if it could be set locally. I suppose maybe it could be a `uuid`?

tests/Feature/Generator/MigrationGeneratorTest.php
```diff
@@ -149,9 +150,17 @@ public function output_also_creates_pivot_table_migration()
         $this->files->expects('put')
             ->with($model_migration, $this->fixture('migrations/belongs-to-many.php'));
 
+        Config::set('blueprint.use_constraints', false);
```
jasonmccreary, line 153: Seems odd that this is set and unset in the same test. I also think it needs to be through `$this->app->config` for Testbench.

tests/TestCase.php
```diff
@@ -11,6 +11,7 @@ protected function getEnvironmentSetUp($app)
         $app['config']->set('blueprint.models_namespace', '');
         $app['config']->set('blueprint.app_path', 'app');
         $app['config']->set('blueprint.generate_phpdocs', false);
+        $app['config']->set('blueprint.use_constraints', true);
```
jasonmccreary, line 14: This should default to `false` like the actual config and turned on in whatever test case you are verifying behavior of foreign key constraints.

src/Generators/MigrationGenerator.php
mikevandiepen: How would that be done? The pivot migration gets generated through the `belongsToMany: Model` option and unfortunately the pivot table gets generated before the secondary table which causes this to always fail.. Am I missing something? 🤔

"The best option would be to create all pivot tables last.,, 

I considered this too but I was to unfamiliar with the codebase to make such a major adjustment, therefore my decision.

src/Generators/MigrationGenerator.php
mikevandiepen: I'll leave this one up to you!

src/Generators/MigrationGenerator.php
jasonmccreary: No worries. Just revert this line for now since so the original test passes and I'll clean it up after merge as it is a related, but separate issue.

src/Generators/MigrationGenerator.php
jasonmccreary: Well, you're changing the default, so I'd like to know why/how it's used.

mikevandiepen: You're right, I didn't quite catch your question the first time. The default value should be reverted and the value of the datatype when `use_constraints` is enabled the type should be `foreignId`.

tests/Feature/Generator/MigrationGeneratorTest.php
mikevandiepen: This decision was made to test the migration when `use_constraints` is enabled and disabled.

jasonmccreary: It would be good to see these in two separate tests. In fact, I would expect you only need one test to demonstration the constraints.

(and another to demonstrate using `foreignId` for Laravel 7+.

src/Generators/MigrationGenerator.php
mikevandiepen: Gave it some thought and I might have come up with a solution. A separate migration could be generated at the end which will solely be to create constraints. For example;
```
        Schema::table('comment_post', function (Blueprint $table) {
            $table->foreign('comment_id')->references('id')->on('comments');
            $table->foreign('post_id')->references('id')->on('posts');
        });
```

mikevandiepen:
@ttomdewit If I'm correct we can ignore the order of the columns in general by implementing the method described [here](https://github.com/laravel-shift/blueprint/pull/141#discussion_r414500475)

I know that isn't the ideal solution with in my opinion is it the best way to make it fail safe.

src/Generators/MigrationGenerator.php
```diff
@@ -156,26 +156,44 @@ protected function buildDefinition(Model $model)
...
-                } else {
-                    $definition .= self::INDENT . '$table->unsignedBigInteger' . "('{$localeColumn}');" . PHP_EOL;
-                    $definition .= self::INDENT . '$table->foreign' . "('{$localeColumn}')->references('{$foreignColumn}')->on('{$foreignTable}');" . PHP_EOL;
+            $definition .= $this->buildConstraint($foreign, $on, $references, $dataType);
+        }
+
+        return trim($definition);
+    }
+
+    protected function buildConstraint(string $foreign, string $on, string $references = 'id', string $dataType = 'unsignedBigInteger', ?string $onDelete = null)
```
jasonmccreary, line 175: What are the other options to `onDelete`?

tests/Feature/Generator/MigrationGeneratorTest.php
jasonmccreary: I still don't understand this line. Why is it needed when it's defaulted to `false` in `TestCase::setUp`?

tests/TestCase.php
mikevandiepen: I'm currently busy with updating the tests at this point, changes will deployed in a little while!

src/Generators/MigrationGenerator.php
mikevandiepen: Default MySQL constraints are `['cascade', 'set null', 'restrict']` and I assumed Laravel had support for all the options, but now apparently they have them through other way, will adjust this.

ttomdewit:
> @ttomdewit If I'm correct we can ignore the order of the models in general by implementing the method described [here](https://github.com/laravel-shift/blueprint/pull/141#discussion_r414500475)
> 
> I know that isn't the ideal solution with in my opinion is it the best way to make it fail safe.

I’m playing with the Draft file and I’m pretty sure I can get the order correctly set on the fly. I’ll give it some attention today, perhaps it’s possible without an additional migration.

src/Generators/MigrationGenerator.php
jasonmccreary: What do you mean "now". Like in Laravel 7 they changed?

mikevandiepen: [This](https://stackoverflow.com/questions/39518162/available-actions-for-onupdate-ondelete-in-laravel-5-x) stack-overflow issue from (5.3) addressed this, I quickly assumed this was the way to go since it seemed rather simple. But doing further research the Laravel (7.x) documentation also addresses this but very subtle; [Laravel Docs](https://laravel.com/docs/7.x/migrations#foreign-key-constraints) 

> You may also specify the desired action for the "on delete" and "on update" properties of the constraint:
```php
$table->foreignId('user_id')
      ->constrained()
      ->onDelete('cascade');
```

mikevandiepen:
The only issue which I haven't been able to resolve is mocking the right Laravel version in the test, can someone maybe steer me in the right direction?

jasonmccreary:
@mikevandiepen thanks for all this work. You can mock the `App` facade [as done here](https://github.com/laravel-shift/blueprint/blob/08ec7c8e32c7f63ab507a97182134b6b3b671537/tests/Feature/Generator/MigrationGeneratorTest.php#L102).

If that's not working, just let me know where you left off and I'll finish this up as it's a good addition.

mikevandiepen:
> 
> 
> @mikevandiepen thanks for all this work. You can mock the `App` facade [as done here](https://github.com/laravel-shift/blueprint/blob/08ec7c8e32c7f63ab507a97182134b6b3b671537/tests/Feature/Generator/MigrationGeneratorTest.php#L102).
> 
> If that's not working, just let me know where you left off and I'll finish this up as it's a good addition.

I have tried this but unfortunately with no prevail, I'm to unfamiliar with the whole code-base to figure out why some things are failing. Also a ton of mockery related Exceptions appear, for example; ```Method version(<No Arguments>) from Mockery_0 should be called
 exactly 1 times but called 5 times.``` 

And for the contributions, I don't mind. I sort of "need" this implementation for my own project so I'm happy to help others and myself out :)

jasonmccreary:
No problem. I'll hack on this over the weekend and get it merged and tagged.

Thanks again!

ttomdewit:
> @ttomdewit If I'm correct we can ignore the order of the models in general by implementing the method described [here](https://github.com/laravel-shift/blueprint/pull/141#discussion_r414500475)
> 
> I know that isn't the ideal solution with in my opinion is it the best way to make it fail safe.

I played with it some more tonight and have a _super_ dirty solution in a test file

```php
public function test_it_sets_correct_model_order_for_foreign_key_migrations(): void
{
    $models = [
        'post' => [
            'user_id' => 'id',
            'published_at' => 'nullable timestamp',
            'timestamps' => 'timestamps',
        ],
        'user' => [
            'name' => 'string',
            'timestamps' => 'timestamps',
        ],
        'comment' => [
            'user_id' => 'id',
            'post_id' => 'id',
            'published_at' => 'nullable timestamp',
            'timestamps' => 'timestamps',
        ],
    ];

    $correctMigrationOrder = [
        'user',
        'post',
        'comment',
    ];

    $dependencies = [];

    foreach ($models as $name => $attributes) {
        $dependencies[$name] = [];
        foreach ($attributes as $column => $attribute) {
            if (Str::endsWith($column, '_id')) {
                $dependencies[$name][] = str_replace('_id', '', $column);
            }
        }
    }

    $numberOfModelsToCreate = count(array_keys($dependencies));
    $availableModels = [];

    $dependants = $dependencies;

    while ($numberOfModelsToCreate) {
        foreach ($dependants as $name => $foreignKeys) {
            if (count(array_intersect($availableModels, $foreignKeys)) === count($foreignKeys)) {
                $availableModels[] = $name;
                unset($dependants[$name]);

                $numberOfModelsToCreate--;
            }
        }
    }

    $this->assertEquals(
        $correctMigrationOrder,
        $availableModels
    );
}
```

Please note I know this code ain't fancy, but I wanted to have a working example to show _it can be done_.

If this doesn't land into this feature I'll port it to my own Blueprint add-on.

ttomdewit:
> @ttomdewit If I'm correct we can ignore the order of the models in general by implementing the method described [here](https://github.com/laravel-shift/blueprint/pull/141#discussion_r414500475)
> 
> I know that isn't the ideal solution with in my opinion is it the best way to make it fail safe.

I played with it some more tonight and have a _super_ dirty solution in a test file

```php
public function test_it_sets_correct_model_order_for_foreign_key_migrations(): void
{
    $models = [
        'post' => [
            'user_id' => 'id',
            'published_at' => 'nullable timestamp',
            'timestamps' => 'timestamps',
        ],
        'user' => [
            'name' => 'string',
            'timestamps' => 'timestamps',
        ],
        'comment' => [
            'user_id' => 'id',
            'post_id' => 'id',
            'published_at' => 'nullable timestamp',
            'timestamps' => 'timestamps',
        ],
    ];

    $correctMigrationOrder = [
        'user',
        'post',
        'comment',
    ];

    $dependencies = [];

    foreach ($models as $name => $attributes) {
        $dependencies[$name] = [];
        foreach ($attributes as $column => $attribute) {
            if (Str::endsWith($column, '_id')) {
                $dependencies[$name][] = str_replace('_id', '', $column);
            }
        }
    }

    $numberOfModelsToCreate = count(array_keys($dependencies));
    $availableModels = [];

    $dependants = $dependencies;

    while ($numberOfModelsToCreate) {
        foreach ($dependants as $name => $foreignKeys) {
            if (count(array_intersect($availableModels, $foreignKeys)) === count($foreignKeys)) {
                $availableModels[] = $name;
                unset($dependants[$name]);

                $numberOfModelsToCreate--;
            }
        }
    }

    $this->assertEquals(
        $correctMigrationOrder,
        $availableModels
    );
}
```

Please note I know this code ain't fancy, but I wanted to have a working example to show _it can be done_.

If this doesn't land into this feature I'll port it to my own Blueprint add-on.

mikevandiepen:
Just a quick note; `StudlyCaps` model names fail to generate proper foreign indexes on tables;

for example: 
**Draft**
```YAML
  StudlyCaps:
    name: String:255
    relationships:
      belongsToMany: Subject
```

Fails to create a constraints since the migration will look like this:
```PHP
Schema::create('studlycaps_subject', function (Blueprint $table) {
    $table->foreignId('studlycaps_id')->constrained('studlycaps', 'id');
    $table->foreignId('subject_id')->constrained('subjects', 'id');
});
``` 

Instead of the desired
```PHP
Schema::create('studly_caps_subject', function (Blueprint $table) {
    $table->foreignId('studly_caps_id')->constrained('studly_caps', 'id');
    $table->foreignId('subject_id')->constrained('subjects', 'id');
});
``` 

I won't have time to address this myself this weekend :/

BuenGenio:
Guys, what's the status of this? I'm desperate to have this one working as it's preventing me from advancing on a project which is otherwise a no-brainer, so I was contemplating on hacking this one myself... Just want to make sure there's no duplication of effort... 
Tanks

jasonmccreary:
It's in limbo. The original PR had multiple changes in it. I'm in the process of wrangling it back together and properly testing it.

It's on the list for today. So I hope to have something later. It'd be good if you'd like to test it or share you draft file.

BuenGenio:
Currently, not only is there no control over whether onCascade() needs to be added at all, but the column definition is added to output before "foreign" even starts getting processed, resulting in invalid column definitions in the resulting migrations. 
e.g. :
The following draft.yaml:
models:
    Bag: 
        id: biginteger unsigned
        client_id: foreign:client
        client_code: foreign:client 
        mawb_id: id:mawb
        station: string:4
        depot: string
        value: enum:LV,MV,HV,MX default:'MX'    
        dim_l: decimal:3,1
        dim_w: decimal:3,1
        dim_h: decimal:3,1
        created_by: integer unsigned 
        timestampstz: true  
... 
Generates the following migration
...
            $table->string('client_id');
            $table->foreignId('client_id')->constrained('clients', 'id')->onDelete('cascade');
            $table->string('client_code');
            $table->foreignId('client_code')->constrained('clients', 'id')->onDelete('cascade');

Not only is the table definition invalid, there is no control over the target column being referenced, nor is there an option to avoid adding any referential actions altogether...

Besides, I personally find Laravel v7 foreignID shorthand way too obscure and would much rather prefer to stick with something more descriptive, explicit and easier to read, like: 
$table->foreign('client_id')->references('id')->on->('clients');

Last but not least, while it would no doubt be cool to have "client_id" automatically be translated to the above, I feel there is still lack of control over referencing non-PK columns in other tables...

With all of the above being said, would everyone happy with the following  slightly modified "foreign" syntax:  

e.g. 
models: 
    Clients: 
        name: string:50
        code: string:10
        ...
    Bags: 
        client_code: foreign:code on:clients
        ...
which would translate to: 
       $table->foreign('client_code')->references('code')->on('clients');

That's, of course, in addition to finding the cleanest possible way of queueing migrations in the right order to have any referenced tables created before they are referenced...

I have the whole evening to work on this if I can get some feedback...

Thanks for one of the most useful packages ever in the meantime, which is just this PR short of a swiss army knife for World Domination :D

jasonmccreary:
Please understand the goal is not to be swiss army knife. You _should_ do some work.

IMO, foreign key constraints are a waste of time. However, I understand some systems like using them. So I want them to be supported. But you could write a thousand lines of code to handle all the variations of migration code generated between keys, tables, columns, references, indexes, cascade, and Laravel versions.

tl;dr; I'm going to finish this PR, but don't expect Blueprint to met all use cases for foreign keys.

jasonmccreary:
@mikevandiepen and @BuenGenio, would appreciate if you pointed your dependency at `dev-mikevandiepen-master` to try out the latest updates.

Please leave comment on #154.

PR #144, opened 2020-04-26
Fix duplicated factories

### LINKED ISSUE #142: Wrong http method in destroy_deletes TestCase
Using the following config:

```
models:
  Post:
    title: string:400
    content: longtext
    published_at: nullable timestamp

controllers:
  Post:
    index:
      query: all
      render: post.index with:posts

    store:
      validate: title, content
      save: post
      send: ReviewNotification to:post.author with:post
      dispatch: SyncMedia with:post
      fire: NewPost with:post
      flash: post.title
      redirect: post.index

    destroy:
      delete: post
      fire: PostDeletedEvent with:post
```

The generated test case `destroy_deletes` is:

```
    /**
     * @test
     */
    public function destroy_deletes()
    {
        $post = factory(Post::class)->create();
        $post = factory(Post::class)->create();

        Event::fake();

        $response = $this->get(route('post.destroy', $post));

        $this->assertDeleted($post);

        Event::assertDispatched(PostDeletedEvent::class, function ($event) use ($post) {
            return $event->post->is($post);
        });
    }
```

As you can see the HTTP method being used is "GET", but the correct method should be "DELETE".

Pr3d4dor:
This Fixes duplicated factories being in the tests.

If the controller method is called 'destroy' and contains the 'delete' comand, the factory will be duplicated because of these parts:

```
if (in_array($name, ['edit', 'update', 'show', 'destroy'])) {
    $setup['data'][] = sprintf('$%s = factory(%s::class)->create();', $variable, $model);
}
```

```
elseif ($statement->operation() === 'delete') {
    $tested_bits |= self::TESTS_DELETE;
    $setup['data'][] = sprintf('$%s = factory(%s::class)->create();', $variable, $model);
    $assertions['generic'][] = sprintf('$this->assertDeleted($%s);', $variable);
}
```

The same will occur if the method is called 'show' and contains 'find' in the blueprint definition for example.

Closes #142

### DIFF
--- a/src/Generators/TestGenerator.php
+++ b/src/Generators/TestGenerator.php
@@ -280,7 +280,11 @@ class TestGenerator implements Generator
                     $assertion = sprintf('$response->assertRedirect(route(\'%s\'', $statement->route());
 
                     if ($statement->data()) {
-                        $assertion .= ', [' . $this->buildParameters($this->data()) . ']';
+                        $parameters = array_map(function ($parameter) {
+                            return '$' . $parameter;
+                        }, $statement->data());
+
+                        $assertion .= ', [' . implode(', ', $parameters) . ']';
                     } elseif (Str::contains($statement->route(), '.')) {
                         [$model, $action] = explode('.', $statement->route());
                         if (in_array($action, ['edit', 'update', 'show', 'destroy'])) {
--- a/src/Generators/TestGenerator.php
+++ b/src/Generators/TestGenerator.php
@@ -335,7 +335,6 @@ class TestGenerator implements Generator
                         $setup['data'][] = sprintf('$%s = factory(%s::class)->create();', $variable, $model);
                     } elseif ($statement->operation() === 'delete') {
                         $tested_bits |= self::TESTS_DELETE;
-                        $setup['data'][] = sprintf('$%s = factory(%s::class)->create();', $variable, $model);
                         $assertions['generic'][] = sprintf('$this->assertDeleted($%s);', $variable);
                     }
                 } elseif ($statement instanceof QueryStatement) {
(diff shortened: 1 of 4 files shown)


jasonmccreary:
I'd like to see a test which backed this up. I understand the change for `delete`, but `show` particularly worries me.

Pr3d4dor:
@jasonmccreary OK, I will add a test for it, if you think there is a better approach to fix this you can close this pull request.

jasonmccreary:
I think what you are fixing is correct. However, I want to add some test coverage for the test generation and I feel the logic you removed was added for convenience. So I'm reluctant to remove it without at least a few tests to demonstrate it needs to be removed.

Pr3d4dor:
@jasonmccreary I removed another line when the operation is delete.

The `['edit', 'update', 'show', 'destroy']` convenience conflicts with specific operations.

The downside of this approach is that find needs to be called explicitly.

I also added a full-crud test for demonstration.

Maybe you could find a better solution and keep the `['edit', 'update', 'show', 'destroy']` convenience.

jasonmccreary:
The more I look at this, the more I consider simply running `array_unique($setup['data'])`.

Consider the following definition:

```yaml
controllers:
  Example:
    action:
        find: SomeModel
        delete: SomeModel
        redirect: some.action with:SomeModel
```

Before, this would generate 3 `SomeModel` model factories. Which is incorrect. But now, I worry generates 0.

This isn't something you need to solve. I think the tests for the test generation was weak. It needs to be strengthened over time.

Pr3d4dor:
@jasonmccreary Do you want me to update this pull request using `array_unique($setup['data'])`?

jasonmccreary:
Part of it feels like that's the simplest solution now that solves the problem without introducing other side effects that we may not know about since this is untested.

Pr3d4dor:
@jasonmccreary I've updated the pull request using `array_unique` to remove the duplication. I've also updated the full-crud-example test.

jasonmccreary:
Thanks for this. There have been a lot of changes to `master` since this around the _build_. So I am going to close this to reopen a rebased version so I can see the build pass.

PR #146, opened 2020-04-27
fix: update regular expression to support windows

### LINKED ISSUE #138: softDeletes not working
Hello there.  
I'm new to here and following README to experience blueprint world.  

error_blueprint_softdeletes_00

When I add **softDeletes** to model in yaml, it throws parse exception like below.  

error_blueprint_softdeletes

I saw related issue [#45](https://github.com/laravel-shift/blueprint/issues/45), seems that this is fixed.  
But no good with shorthands for me. It works good with valid yaml syntax like **softDeletes: true**.  

Is this feature(shorthands) unavailable now?

nathane:
> `$` By default, the match must occur at the end of the string or before `\n` at the end of the string; in multiline mode, it must occur before the end of the line or before `\n` at the end of the line.
> https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-language-quick-reference

closes #138

### DIFF
--- a/src/Blueprint.php
+++ b/src/Blueprint.php
@@ -26,20 +26,22 @@ class Blueprint
 
     public function parse($content)
     {
+        $content = str_replace(["\r\n", "\r"], "\n", $content);
+
         $content = preg_replace_callback('/^(\s+)(id|timestamps(Tz)?|softDeletes(Tz)?)$/mi', function ($matches) {
-            return $matches[1] . strtolower($matches[2]) . ': ' . $matches[2];
+            return $matches[1].strtolower($matches[2]).': '.$matches[2];
         }, $content);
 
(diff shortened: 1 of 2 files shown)


jasonmccreary:
While I understand since I followed the issue, but it seems rather cryptic.

I'd rather use `\R` instead of `\r?$` or normalize `$contents` beforehand in a similar way.

We should also add a parser test where we pass in a string with Windows line endings to verify this change.

nathane:
> I'd rather use `\R` instead of `\r?$` or normalize `$contents` beforehand in a similar way.

**I have chosen to normalize `$contents` with `str_replace(["\r\n", "\r"], "\n", $content);`**

> We should also add a parser test where we pass in a string with Windows line endings to verify this change.

**The test gives a false positive on newer versions of macOS. It must pass on Mac (OS 9 and below) and Windows OS for confidence.** (unless I missed something)

jasonmccreary:
Thanks.

PR #149, opened 2020-04-27
feat: add php-cs-fixer github actions

nathane:
This PR includes `.php_cs` generated by @laravel-shift to allow contributors to run fix locally.

Install globally with Composer `composer global require friendsofphp/php-cs-fixer`.

Usage: `php-cs-fixer fix`


jasonmccreary:
At some point StyleCI was doing this against `master`. But it's not anymore. Let me check that first as I don't want to rely solely on users having to run `php-cs-fixer`.

nathane:
Sure thing. We can also use github actions to run `php-cs-fixer` and commit the changes.

jasonmccreary:
I appreciate this. But again, StyleCI is already built into the build process. I just need to add an option to have it run automatically.

PR #153, opened 2020-04-28
fix: tests on windows

nathane:
This PR will fix all the tests that were failing on windows.

**P.S. If you decide to merge this PR remove `continuous-integration.yml`. It's only included it in this PR to show that the tests are now passing.**

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,0 +1,64 @@
+name: Continuous Integration
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: "0 0 * * *"
+
+env:
+  COMPOSER_FLAGS: "--ansi --no-interaction --no-progress --no-suggest --prefer-dist -o -n"
+
+jobs:
+  tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version: [7.4]
+        laravel-version: [6, 7]
+        os: [windows-latest]
+        dependencies: [locked]
+        experimental: [false]
+
+    name: PHP ${{ matrix.php-version }} - L${{ matrix.laravel-version }} - ${{ matrix.os }}
+
+    runs-on: ${{ matrix.os }}
+
+    continue-on-error: ${{ matrix.experimental }}
+
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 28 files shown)


jasonmccreary:
I don't understand the need for this. All tests are created on a Unix line-ending environment. Why would every stub need to be normalized?

nathane:
>`PHP_EOL` represents the end of line character for the current system.

nathane:
https://github.com/php/php-src/blob/7d1c8064912da100be0106774b9ba08ec1a88ee6/main/php.h#L58-L76

jasonmccreary:
I know.

Still doesn't answer the question why we need to normalize in every generator. If you simply normalized the `stub` macro and the `fixture` macro, that should be sufficient so long as `PHP_EOL` is used everywhere else...

nathane:
Because each generator's custom `populateStub` method is returning a string that includes `PHP_EOL`.

jasonmccreary:
Maybe I'm missing something, but we _want_ the line endings to be whatever `PHP_EOL` returns.

The issue on Windows, to my understanding, is that some pieces of this file (e.g. the stubs) a Unix line endings.

So, it would seem if we normalize the stubs to use `PHP_EOL` and use `PHP_EOL` everywhere else (e.g. in the generators), all is good.

I would expect this to be a 3 line change (in the macros). Not 28 files.

nathane:
the new `stubs(string $path)` method only normalizes the pre-existing code that is shipped with blueprint, not the code that is being generated (which included `PHP_EOL`).

I did the same thing you mentioned but it didn't work, which is why I had to went threw each generator to normalize the output.

https://github.com/nathane/blueprint/runs/624527207?check_suite_focus=true#step:9:1

jasonmccreary:
Alright. I'll take a closer look. As it stands, I don't want every generator to have to know "how" to normalize the files.

Maybe there needs to be another `files` macro instead of `put`.

jasonmccreary:
Upon farther review. I think the root cause is the directory separator. From what I can tell, either the tests or the underlying filesystem is normalizing the line endings.

I'm working on a separate branch where I normalize all paths to Unix style as it seems Windows accepts both.

PR #154, opened 2020-04-28
Generate foreign key constraints

### LINKED ISSUE #137: Use foreignId for belongsTo relationships in Laravel 7

jasonmccreary:
This adds support for foreign key constraints. While syntax existed for defining `foreign` keys, it was incomplete.

So there's quite a lot in this PR. More than I might have originally added to Blueprint. As such, future enhancements on this will be limited as I believe it would be faster for you to make any additional adjustments by hand.

With that said, this PR includes:

- A new `use_constraints` config option to always generate foreign key constraints for `id` and `uuid` columns where the column name follows Laravel conventions. This option is disabled by default.
- if you wish to generate foreign key constraints for some fields, but not all, you may also add the `foreign` modifier to your column definition to generate constraints.
- Improved granularity when using the `foreign` syntax to supply the foreign table or model reference, with or without a column name. For example, `foreign:country`, `foreign:Country`, `foreign:countries`, and `foreign:countries.id` are all examples of acceptable formats. However, this should only be need when you use unconventional names or foreign keys.
- Support for Laravel 7 streamlined `foreignId`, `constrained`, and `cascadeOnDelete` methods.

---
This supersedes #141 and closes #137


BuenGenio:
@mikevandiepen Mike, it's now generating the long-form references, which is a lot easier to read. 
I noticed ->onCascade('delete') is now hard coded inside buildForeignKey() - previously it was possible to decide whether to add it or not through the 3rd ?onDelete = null param (it was still always getting set for Laravel 7.x)... 

For me the main two issues are these: 
1) Do not add the column type definition if we're adding ->foreign('column'), as it results in duplication of column names and breaks the migration; and 
2) Order in which migrations are generated, based on foreign key precedence. 

Have you any plans to address these and any possible solutions in mind, or do you need help?

@jasonmccreary What you describe is absolutely spot on - there is no way around putting in proper and thorough thinking into database design, no tool is going to replace that and I had already put in over two days into my migrations design before I accidentally bumping into Blueprint. The problem I am hoping Blueprint will help me solve is keeping my models, their attributes and their types in sync and consistent both when I'm prototyping and when I need to add, remove or update some attributes during development...

jasonmccreary:
@BuenGenio, this PR is mine. Whatever changes to @mikevandiepen's original code I made explicitly. Let me know what you feel has issues.

BuenGenio:
Here are my two main issues. Consider the following draft.yaml: ```
models:
    Bag:
        id: biginteger unsigned
        client_id: id:client foreign
        client_code: foreign:clients.code (or code:clients foreign; or simply foreign)
        mawb_id: foreign:mawb
        station: string:4
        depot: string
        value: enum:LV,MV,HV,MX default:'MX'
        dim_l: decimal:3,1
        dim_w: decimal:3,1
        dim_h: decimal:3,1
        created_by: integer unsigned
        timestampstz: true

```
Produces the following output: ```
Schema::create('bags', function (Blueprint $table) {
            $table->unsignedBigInteger('id');
            $table->foreignId('client_id')->constrained()->cascadeOnDelete();
            $table->string('client_code'); <-- this should be excluded since we have the line below
            $table->foreign('client_code')->references('code')->on('clients')->onDelete('cascade');
            $table->string('mawb_id');
            $table->foreign('mawb_id')->references('id')->on('mawbs')->onDelete('cascade');
...
```
The second issue is that the table `clients` needs to be created first, or the migration will fail due to non-existent foreign key reference.

jasonmccreary:
@BuenGenio, first, my understanding is `foreign()` simply creates the index. You still need to create the column.

Second, Blueprint is not responsible for order or operations. That is on the developer, just as it would be normally.

BuenGenio:
OK, fair enough Re: order of model definition, didn't think of that till today :)

Regarding foreign Laravel automatically creates the column definition with the correct definition of the column being referenced, so any explicit column definition prior to ->foreign needs to be dropped from output. 

Also, just noticed a new issue:
``` 
 Flight:
        flight: string:8
        carrier: string:3
        departure: datetimetz
        arrival: datetimetz
        origin: string:3
        destination: string:3
  ***  parentOf: id:flight ***
``` 
produces 
```
...
$table->foreignId('parentOf')->constrained('parentOfs', 'parentOf')->cascadeOnDelete());
```
the very last closing bracket causing a syntax error as it shouldn't be there...

Also, I think there's an error with ->foreignId (which is called if a field ends in _id) - should be "flights" not "parentOfs"

jasonmccreary:
I'm having a hard time following your examples as every time you provide a different one and only pieces of the code.

I'll test my own example to see if these are indeed duplicated, as I don't believe `foreign` does anything more than make the `index`, while `foreignId` (Laravel 7) does make both the column and index (for integer columns).

BuenGenio:
OK, here's my draft.yaml: 
https://pastebin.ubuntu.com/p/G6NNg2tYpH/

here's the output: 
```
./artisan migrate

   ParseError 

  syntax error, unexpected ')'

  at database/migrations/2020_04_28_220826_create_flights_table.php:24
    20|             $table->dateTimeTz('departure');
    21|             $table->dateTimeTz('arrival');
    22|             $table->string('origin', 3);
    23|             $table->string('destination', 3);
  > 24|             $table->foreignId('parentOf')->constrained('parentOfs', 'parentOf')->cascadeOnDelete());
    25|             $table->timestamps();
    26|         });
    27|     }
    28| 

      +19 vendor frames 
  20  artisan:37
      Illuminate\Foundation\Console\Kernel::handle()
```
and after fixing the syntax error in flights migration: 
```
Migrating: 2020_04_28_220825_create_bags_table

   Illuminate\Database\QueryException 

  SQLSTATE[HY000]: General error: 1005 Can't create table `edms3`.`bags` (errno: 150 "Foreign key constraint is incorrectly formed") (SQL: alter table `bags` add constraint `bags_client_id_foreign` foreign key (`client_id`) references `id` (`clients`) on delete cascade)

  at vendor/laravel/framework/src/Illuminate/Database/Connection.php:671
    667|         // If an exception occurs when attempting to run a query, we'll format the error
    668|         // message to include the bindings with SQL, which will make this exception a
    669|         // lot more helpful to the developer instead of just the database's errors.
    670|         catch (Exception $e) {
  > 671|             throw new QueryException(
    672|                 $query, $this->prepareBindings($bindings), $e
    673|             );
    674|         }
    675| 

      +11 vendor frames 

```

BuenGenio:
Yes, about ->foreign and the column definition, you're right about column type being required, though foreign() is redundant in that case. 

Here's the correct one liner which the creating a column and adding a foreign reference should produce:
```
$table->unsignedBigInteger('mawb_id')->references('mawb_id')->on('mawbs');
```

BuenGenio:
> Yes, about ->foreign and the column definition, you're right about column type being required, though foreign() is redundant in that case.
> 
> Here's the correct one liner which the creating a column and adding a foreign reference should produce:
> 
> ```
> $table->unsignedBigInteger('mawb_id')->references('mawb_id')->on('mawbs');
> ```

also, `mawb_id` is the name of the actual PK in `mawbs`
Note that this is the original migration I did by hand a last week, so `->references('mawb_id')` is what I actually have, as Blueprint would have yielded just `id`

jasonmccreary:
Both bits of code to the same thing. I wouldn't call it redundant per se, it's just that `references` is a shorthand.

As far as the `parentOf: id:flight`, that is indeed a boundary case not covered yet where the referencing model could be inferred from the `id` attribute.

jasonmccreary:
I'm going to go ahead and merge this but not tag it. That way others can continue to try against `dev-master`.

I'll also open an issue for the `parentOf: id:flight` definition. Maybe it's something you can hack on.

BuenGenio:
Roger! Great, I definitely think that area could do with a poke, along with testing whether it's the best way to do multi-legged flights 




On Wed, Apr 29, 2020 at 4:19 PM +0300, "Jason McCreary" <notifications@github.com> wrote:















I'm going to go ahead and merge this but not tag it. That way others can continue to try against dev-master.


I'll also open an issue for the parentOf: id:flight definition. Maybe it's something you can hack on.

jasonmccreary:
Okay. I would encourage you to stick to the issues or open separate PRs. I'm not going to tweak this much, only fix any bugs with the existing syntax.

BuenGenio:
Just merged master into my fork, which is required as dev-master in composer. 
I can't replicate the syntax error bug any more...  Has this been fixed?
I'll work on #156 for now.

BuenGenio:
> OK, here's my draft.yaml:
> https://pastebin.ubuntu.com/p/G6NNg2tYpH/
> 
> here's the output:
> 
> ```
> ./artisan migrate
> 
>    ParseError 
> 
>   syntax error, unexpected ')'
> 
>   at database/migrations/2020_04_28_220826_create_flights_table.php:24
>     20|             $table->dateTimeTz('departure');
>     21|             $table->dateTimeTz('arrival');
>     22|             $table->string('origin', 3);
>     23|             $table->string('destination', 3);
>   > 24|             $table->foreignId('parentOf')->constrained('parentOfs', 'parentOf')->cascadeOnDelete());
>     25|             $table->timestamps();
>     26|         });
>     27|     }
>     28| 
> 
>       +19 vendor frames 
>   20  artisan:37
>       Illuminate\Foundation\Console\Kernel::handle()
> ```
> 
> and after fixing the syntax error in flights migration:
> 
> ```
> Migrating: 2020_04_28_220825_create_bags_table
> 
>    Illuminate\Database\QueryException 
> 
>   SQLSTATE[HY000]: General error: 1005 Can't create table `edms3`.`bags` (errno: 150 "Foreign key constraint is incorrectly formed") (SQL: alter table `bags` add constraint `bags_client_id_foreign` foreign key (`client_id`) references `id` (`clients`) on delete cascade)
> 
>   at vendor/laravel/framework/src/Illuminate/Database/Connection.php:671
>     667|         // If an exception occurs when attempting to run a query, we'll format the error
>     668|         // message to include the bindings with SQL, which will make this exception a
>     669|         // lot more helpful to the developer instead of just the database's errors.
>     670|         catch (Exception $e) {
>   > 671|             throw new QueryException(
>     672|                 $query, $this->prepareBindings($bindings), $e
>     673|             );
>     674|         }
>     675| 
> 
>       +11 vendor frames 
> ```

This one

jasonmccreary:
You must have pulled against the feature branch at some point where the code was incomplete. To your point, doesn't seem to be an issue in `master`, after merging.

In the future, please open new issues to report existing bugs, questions, or ideas. Do not comment on closed/merged PRs.

BuenGenio:
Copy.

PR #159, opened 2020-04-29
Accept both draft.yaml and draft.yml file

### LINKED ISSUE #157: Accept both draft.yaml and draft.yml file
Currently `blueprint:build` will automatically load a `draft.yaml` if no draft file is passed to the command.

For convenience, Blueprint should also look for a `draft.yml` file to allow for the alternative file extension.

ericvanjohnson:
Not the most eloquent solution but it should achieve the result you are looking for. 

---
Closes #157

### DIFF
--- a/src/Commands/BlueprintCommand.php
+++ b/src/Commands/BlueprintCommand.php
@@ -16,7 +16,7 @@ class BlueprintCommand extends Command
      *
      * @var string
      */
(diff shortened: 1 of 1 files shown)


jasonmccreary:
Thanks man. Did a little "boy scouting", but looks good. I'll backfill some tests for it moving forward.

PR #169, opened 2020-05-01
Fix/duplicated pivot table

### LINKED ISSUE #167: belongsToMany generates duplicate migrations for pivot tables
Given the following contents of `draft.yaml`:
```yaml
models:
  Company:
    name: string
    relationships:
      belongsToMany: Person
  Person:
    name: string
    relationships:
      belongsToMany: Company
```
Generates:
```sh
- database/migrations/2020_05_01_144405_create_companies_table.php
- database/migrations/2020_05_01_144405_create_company_person_table.php
- database/migrations/2020_05_01_144406_create_people_table.php
- database/migrations/2020_05_01_144406_create_company_person_table.php
- app/Company.php
- app/Person.php
- database/factories/CompanyFactory.php
- database/factories/PersonFactory.php
```
Observe that there are 2 migrations for the same pivot table.

Inspecting the two, we find a declaration with the same Class name: `CreateCompanyPersonTable`. This, of course, is a problem.

```sh
PHP Fatal error:  Cannot declare class CreateCompanyPersonTable, because the name is already in use in /Users/joostjacobs/Code/blueprint/database/migrations/2020_05_01_144406_create_company_person_table.php on line 7
```

Inspecting the two migrations, the only observable difference is the order in which the columns appear in the schema:

Inside `database/migrations/2020_05_01_144405_create_company_person_table.php
`:
```php
    public function up()
    {
        Schema::create('company_person', function (Blueprint $table) {
            $table->unsignedBigInteger('company_id');
            $table->unsignedBigInteger('person_id');
        });
    }
```

Inside `database/migrations/2020_05_01_144406_create_company_person_table.php`:
```php
    public function up()
    {
        Schema::create('company_person', function (Blueprint $table) {
            $table->unsignedBigInteger('person_id');
            $table->unsignedBigInteger('company_id');
        });
    }
```

Pr3d4dor:
Closes #167


src/Generators/MigrationGenerator.php
```diff
@@ -53,18 +53,18 @@ public function output(array $tree): array
             if (!empty($model->pivotTables())) {
                 foreach ($model->pivotTables() as $pivotSegments) {
                     $pivotTable = $this->getPivotTableName($pivotSegments);
-                    if (isset($created_pivot_tables[$pivotTable])) {
```
jasonmccreary, line 56: I'd rather leave this code in the loop and update the `isset` to `in_array`. This prevents a duplicate loop and is the simplest change.

Pr3d4dor: @jasonmccreary I tried that, but with `in_array` the migration order is incorret.

With `in_array` the migrations are:
```
[
    0 => "database/migrations/2020_05_01_191229_create_companies_table.php"
    1 => "database/migrations/2020_05_01_191229_create_company_person_table.php"
    2 => "database/migrations/2020_05_01_191230_create_people_table.php"
]
```

This way, migrations are:
```
[
    0 => "database/migrations/2020_05_01_191229_create_companies_table.php"
    1 => "database/migrations/2020_05_01_191230_create_people_table.php"
    2 => "database/migrations/2020_05_01_191230_create_company_person_table.php"
]
```

jasonmccreary: I see. And the tests confirm the later timestamp, so I guess that's good.

PR #173, opened 2020-05-02
Allow configuration to use `$guarded = [];`

### LINKED ISSUE #148: Allow configuration to use `$guarded = [];`
This as probably come up in conversation somewhere, but I was unable to find anything.
First of all what a great package @jasonmccreary works a bloody treat, love the fact it creates factories, lot of time saving there in itself.

Anyway, each model is created with the `$fillable` attribute, it would be cool if you could specify what you wish to use i/e `$fillable` or `$guarded` with an empty array.

Just a thought, maybe something to add I dunno...

Wowenko:
Use `$guarded` instead of `$fillable` when configuration option is set.

---
Closes #148

### DIFF
--- a/config/blueprint.php
+++ b/config/blueprint.php
@@ -69,4 +69,15 @@ return [
     'use_constraints' => false,
 
     'fake_nullables' => true,
+
+    /*
+    |--------------------------------------------------------------------------
+    | Models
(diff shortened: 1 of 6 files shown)


jasonmccreary:
Very nice. Thank you for this. I will review and merge shortly.

PR #174, opened 2020-05-02
feat: between rule for numeric types

### LINKED ISSUE #24: Add `between` rule for decimal or float columns
When a "decimal" column is defined with _precision_ and _decimal_ attributes it would be possible to add a `between` rule.

For example, given the definition:

```yaml
total: decimal:10,2
```

Blueprint could add `between` to generate the following rule:

```php
'decimal' => 'required|numeric|between:0,999999.99'
```

Pr3d4dor:
- Between rule for numeric types.

Closes #24


src/Translators/Rules.php
```diff
@@ -80,6 +80,18 @@ public static function fromColumn(string $context, Column $column)
             array_push($rules, 'unique:' . $context . ',' . $column->name());
         }
 
+        if (Str::contains($column->dataType(), 'decimal')) {
```
jasonmccreary, line 83: These could all be refactored to `in_array` as the `dataType` will always been one of those.

src/Translators/Rules.php
```diff
@@ -94,4 +106,33 @@ private static function overrideStringRuleForSpecialNames($name)
 
         return 'string';
     }
+
+
+    private static function betweenRuleForNumericTypes(Column $column, $numericType)
+    {
+        $defaultPrecisionAndScale = [8, 2];
```
jasonmccreary, line 113: If there is no _precision_ or _scale_, I wouldn't want to add the `between` rule.

Pr3d4dor:
@jasonmccreary I've already removed the default values, but in order to use `in_array` the definition needs to be `unsignedDecimal` for example and `unsigned decimal` would no be possible in this way.

What do you want me to do?

jasonmccreary:
@Pr3d4dor, maybe I've forgotten now, but I thought the root dataType was always `decimal` and the migration optimized the rest.

Either way, it seems to me you only need to know if the data type _contains_ any of the _numeric_ types and specifies a _value_. If so, then add the `between` rule.

My goal is always to write as little code as possible. Not one liners necessarily. But simple, readable code.

Give it one more review to see if you see anything. Then I will merge it.

Pr3d4dor:
@jasonmccreary Ok, I'll take a look and update the pull request if necessary.

Pr3d4dor:
DECIMAL, FLOAT and DOUBLE aren't the same thing:
https://dev.mysql.com/doc/refman/8.0/en/numeric-types.html

Laravel Code:
```
/**
 * Create a new float column on the table.
 *
 * @param  string  $column
 * @param  int  $total
 * @param  int  $places
 * @param  bool  $unsigned
 * @return \Illuminate\Database\Schema\ColumnDefinition
 */
public function float($column, $total = 8, $places = 2, $unsigned = false)
{
    return $this->addColumn('float', $column, compact('total', 'places', 'unsigned'));
}

/**
 * Create a new double column on the table.
 *
 * @param  string  $column
 * @param  int|null  $total
 * @param  int|null  $places
 * @param  bool  $unsigned
 * @return \Illuminate\Database\Schema\ColumnDefinition
 */
public function double($column, $total = null, $places = null, $unsigned = false)
{
    return $this->addColumn('double', $column, compact('total', 'places', 'unsigned'));
}

/**
 * Create a new decimal column on the table.
 *
 * @param  string  $column
 * @param  int  $total
 * @param  int  $places
 * @param  bool  $unsigned
 * @return \Illuminate\Database\Schema\ColumnDefinition
 */
public function decimal($column, $total = 8, $places = 2, $unsigned = false)
{
    return $this->addColumn('decimal', $column, compact('total', 'places', 'unsigned'));
}
```

jasonmccreary:
What's you point?

Pr3d4dor:
@jasonmccreary Just to verify if `decimal` was the root dataType or not like you said.

jasonmccreary:
No worries. I'll finish this up. It'll be faster for me to just make the changes.

On a side note, I've noticed your last few PRs have a bunch of "merge master" commits. You can avoid this and have a cleaner PR by first pulling in the latest changes from Blueprint into your local `master`. Then create your feature branch.

Issue #172, opened 2020-05-02
Support model reference in `validate` statement to validate all columns


axit-joost:
As mentioned in #171 

Given the following contents of `draft.yaml`:
```yaml
models:
  Certificate:
    name: string
    reference: string
    document: string
    expiry_date: date
    remarks: nullable text

controllers:
  Certificate:
    resource: api
```

Where we use a `resource: api` to generate an REST API Controller.

Running `blueprint:build` will yield:
```sh
- database/migrations/2020_05_02_092122_create_certificates_table.php
- app/Certificate.php
- database/factories/CertificateFactory.php
- app/Http/Controllers/CertificateController.php
- app/Http/Requests/CertificateStoreRequest.php
- app/Http/Requests/CertificateUpdateRequest.php
- app/Http/Resources/CertificateCollection.php
- app/Http/Resources/Certificate.php
- tests/Feature/Http/Controllers/CertificateControllerTest.php
```

If we inspect the StoreRequest and UpdateRequest, and particularly the `rules()` within, we will find the following bogus rule:

```php
    public function rules()
    {
        return [
            'certificate' => 'required',
        ];
    }
```

The expectation here is that the rules should match/correspond with the model's definition.

The generated tests also fail because of this:

```sh
❯ phpunit                                                                                                                                                                                                                         28%
PHPUnit 8.5.4 by Sebastian Bergmann and contributors.

..R.FR.R.                                                           9 / 9 (100%)

Time: 326 ms, Memory: 30.00 MB

There was 1 failure:

1) Tests\Feature\Http\Controllers\CertificateControllerTest::store_saves
Failed asserting that actual size 0 matches expected size 1.

/Users/joostjacobs/Code/blueprint/tests/Feature/Http/Controllers/CertificateControllerTest.php:55
```

jasonmccreary:
I agree. Seems like the `resource` shorthand generates a statement of `validate: certificate`. Which leads more to an unexpected syntax for the `validate` statement.

I think the appropriate solution would be to allow the `validate` statement to accept a _model reference_. In this case, `validate: Certificate`.

Upon doing so, this will attempt to find the model of that name and generate the appropriate validation for all columns.

jasonmccreary:
@Pr3d4dor, I believe you referenced this, but the work you did wasn't on this issue directly, was it?

Care to take a stab at it?

Pr3d4dor:
[...]

PR #181, opened 2020-05-04
fix: controller/resource generation with PascalCase definition

### LINKED ISSUE #171: Problems generating controllers/resources in PascalCase
Seems to be a problem with case-conversion.

Given the contents of the following `draft.yaml`:
```yaml
models:
  Certificate:
    name: string
    certificate_type_id: id
    reference: string
    document: string
    expiry_date: date
    remarks: nullable text
  CertificateType:
    name: string
    relationships:
      hasMany: Certificate

controllers:
  Certificate:
    resource: api
  CertificateType:
    resource: api
```

Running `blueprint:build` will throw the following error:

> TypeError: Argument 1 passed to Blueprint\Generators\Statements\ResourceGenerator::visibleColumns() must be an instance of Blueprint\Models\Model, null given, called in /Users/joostjacobs/Code/blueprint/vendor/laravel-shift/blueprint/src/Generators/Statements/ResourceGenerator.php on line 98

A `.blueprint` file is not created due to this crash.

It ultimately crashed upon trying to create a Resource in `app/Http/Resources/CertificateType.php`.

Examining the partially generated code, the following issues surface:

1. In the generated `app/Http/Controllers/CertificateTypeController.php`, we see two imports, one with the correct PascalCase, one with only the first letter capitalized:

```php
<?php

namespace App\Http\Controllers;

use App\CertificateType;
use App\Certificatetype;
```

2. The generated DocBlocks as well as the method body, refer to an incorrectly cased Collection and Resource: `CertificatetypeCollection` should be `CertificateTypeCollection` and the Resource respectively should not be `CertificatetypeResource` but `CertificateTypeResource`.

```php
    /**
     * @param \Illuminate\Http\Request $request
     * @return \App\Http\Resources\CertificatetypeCollection
     */
    public function index(Request $request)
    {
        $certificatetypes = Certificatetype::all();

        return new CertificatetypeCollection($certificatetypes);
    }

    /**
     * @param \App\Http\Requests\CertificateTypeStoreRequest $request
     * @return \App\Http\Resources\Certificatetype
     */
    public function store(CertificateTypeStoreRequest $request)
    {
        $certificatetype = Certificatetype::create($request->all());

        return new CertificatetypeResource($certificatetype);
    }
```

3. While the method signatures for `show`, `update`, and `destroy` show a correct model binding of `CertificateType $certificateType`, with the variable properly camelCased, the method body uses an all lowercase variable name:

```php
    /**
     * @param \Illuminate\Http\Request $request
     * @param \App\CertificateType $certificateType
     * @return \App\Http\Resources\Certificatetype
     */
    public function show(Request $request, CertificateType $certificateType)
    {
        return new CertificatetypeResource($certificatetype);
    }
```

4. All the generated `Request` classes for validation are incorrect. They do not feature any of the model's columns, and generates a plain wrong rule instead. I think this is unrelated to the PascalCase problem, but it should be raised as an entirely new issue all together:

```php
    /**
     * Get the validation rules that apply to the request.
     *
     * @return array
     */
    public function rules()
    {
        return [
            'certificatetype' => 'required',
        ];
    }
```

I'll raise a separate issue for this.

Pr3d4dor:
I haven't implemented the tests for the generated resources because of the problem in the issue #172

Also because of the issue #172 the generated tests looks incomplete.

Closes #171

### DIFF
--- a/src/Generators/TestGenerator.php
+++ b/src/Generators/TestGenerator.php
@@ -333,7 +333,7 @@ class TestGenerator implements Generator
                             $assertions['sanity'][] = '$this->assertCount(1, $' . $plural . ');';
                             $assertions['sanity'][] = sprintf('$%s = $%s->first();', $variable, $plural);
                         } else {
-                            $assertions['generic'][] = '$this->assertDatabaseHas(' . Str::lower(Str::plural($model)) . ', [ /* ... */ ]);';
+                            $assertions['generic'][] = '$this->assertDatabaseHas(' . Str::camel(Str::plural($model)) . ', [ /* ... */ ]);';
                         }
                     } elseif ($statement->operation() === 'find') {
                         $setup['data'][] = sprintf('$%s = factory(%s::class)->create();', $variable, $model);
@@ -351,7 +351,7 @@ class TestGenerator implements Generator
                 }
             }
 
-            $call = sprintf('$response = $this->%s(route(\'%s.%s\'', $this->httpMethodForAction($name), Str::lower($context), $name);
+            $call = sprintf('$response = $this->%s(route(\'%s.%s\'', $this->httpMethodForAction($name), Str::kebab($context), $name);
 
             if (in_array($name, ['edit', 'update', 'show', 'destroy'])) {
                 $call .= ', $' . Str::camel($context);
--- a/src/Lexers/ControllerLexer.php
+++ b/src/Lexers/ControllerLexer.php
@@ -66,7 +66,7 @@ class ControllerLexer implements Lexer
 
                         return str_replace(
                             ['[singular]', '[plural]'],
-                            [Str::lower($model), Str::lower(Str::plural($model))],
+                            [Str::camel($model), Str::camel(Str::plural($model))],
                             $statement
                         );
                     }),
--- a/src/Models/Statements/EloquentStatement.php
+++ b/src/Models/Statements/EloquentStatement.php
@@ -40,12 +40,12 @@ class EloquentStatement
 
         if ($this->operation() == 'save') {
             if ($context === 'store') {
-                $code = "$" . Str::lower($model);
+                $code = "$" . Str::camel($model);
                 $code .= ' = ';
                 $code .= $model;
                 $code .= '::create($request->all());';
             } else {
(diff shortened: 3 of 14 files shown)


jasonmccreary:
I'm not sure I understand what/how this fixes something. It'd be nice to know what you found the problem. Otherwise, it looks like this just changes everything to `kabab` case.

Pr3d4dor:
@jasonmccreary I was waiting you comment, to point some things out.

The root of the problem is this piece of code:
```
private function generateResourceTokens(Controller $controller, array $methods)
    {
        return collect($this->resourceTokens())
            ->filter(function ($statements, $method) use ($methods) {
                return in_array($method, $methods);
            })
            ->mapWithKeys(function ($statements, $method) use ($controller) {
                return [
                    str_replace('api.', '', $method) => collect($statements)->map(function ($statement) use ($controller) {
                        $model = Str::singular($controller->prefix());

                        return str_replace(
                            ['[singular]', '[plural]'],
                            [Str::lower($model), Str::lower(Str::plural($model))],
                            $statement
                        );
                    }),
                ];
            })
            ->toArray();
    }
```

Because its using `lower`, the string generated is for example:
```
....
'index' => [
    'query' => 'all:certificatetypes',
    'render' => 'certificatetype.index with certificatetypes',
],
....
```

In the `ResourceGenerator` on the method `buildData` it uses the method `modelForContext` to find the correspondent model. The model is not found because it is generated using CamelCase format.

Changing it to camelCase and fixes the problem.

After that, I noted the same problem in other places using `Str::lower` and searched all usages of `Str::lower` and replaced then with `Str::camel`.

This line needs to be in kebab case, because the generated routes are in kebab case:
```
$call = sprintf('$response = $this->%s(route(\'%s.%s\'', $this->httpMethodForAction($name), Str::kebab($context), $name);
```

```
Route::resource('certificate', 'CertificateController')->except('create', 'edit');

Route::resource('certificate-type', 'CertificateTypeController')->except('create', 'edit');
```

jasonmccreary:
Thanks for the fuller explanation. In the future, you should put your findings in the PR description, not wait for a comment.

I'll review this once more and probably merge, then add some tweaks as I worry this is an issue in some other places.

Issue #168, opened 2020-05-01
php artisan blueprint:init


ProjektGopher:
Touches draft.yaml and fills with stub. Possibly run blueprint:trace as well?

jasonmccreary:
Could be cool. Although I'd want to give it a name that aligned with the architectural vocabulary I'm using.

Maybe `blueprint:new` or `blueprint:start` or ...

ProjektGopher:
That makes sense to me. I'll try to whip something together when I've got some time this weekend. Do I fork master for that, or is there a different branch I should use?

jasonmccreary:
Cool. Yeah, Branch off `master`. Follow whatever pattern exists in the `ServiceProvider` to register a command.

Don't worry about _perfection_. Just get it functional. I'll review and tidy up the PR as necessary.

Wowenko:
Hi @jasonmccreary I have tackled this but I'm having problems with the tests. Do you think is worth it to pull something like VFS to test the file creation?

jasonmccreary:
Nope. You can swap the `File` facade and test that way. Go ahead and push what you have. I can update the test.

PR #183, opened 2020-05-05
php artisan blueprint:new

Wowenko:
This is a WIP for #168.

### DIFF
--- a/src/BlueprintServiceProvider.php
+++ b/src/BlueprintServiceProvider.php
@@ -4,6 +4,7 @@ namespace Blueprint;
 
 use Blueprint\Commands\BlueprintCommand;
 use Blueprint\Commands\EraseCommand;
+use Blueprint\Commands\NewCommand;
 use Blueprint\Commands\TraceCommand;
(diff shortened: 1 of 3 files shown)


ProjektGopher:
I think your it_creates_an_empty_draft_file() test would fully create the file during testing wouldn't it? If we use 

$file = \Mockery::mock(Filesystem::class);
        $file->expects('put')
        ->with('draft.yaml');

It would test it without creating the file

ProjektGopher:
I hadn't realized someone else was working on the same feature I was, but in my PR #184 the blueprint:start command also fills the draft.yaml file with stubs/draft.stub. I couldn't figure out the testing though

jasonmccreary:
@Wowenko thanks for this. I am going to close it though to avoid confusion and mark the issue as "in progress".

Feel free to point @ProjektGopher in the right direction as far as tests. But his implementation seems just slightly farther along.

Wowenko:
No problem I'm here to help an learn.

PR #184, opened 2020-05-05
Add blueprint:start command

ProjektGopher:
Once tests are written, will close issue #168

### DIFF
--- a/src/Commands/StartCommand.php
+++ b/src/Commands/StartCommand.php
@@ -54,6 +54,6 @@ class StartCommand extends Command
     
         $this->info('Created example draft.yaml file in project root');
 
-        (new TraceCommand(new Filesystem))->handle();
+        $this->call('blueprint:trace');
     }
 }
--- a/src/Commands/StartCommand.php
+++ b/src/Commands/StartCommand.php
@@ -44,15 +44,15 @@ class StartCommand extends Command
     {
         if ($this->files->exists('draft.yaml')) {
             $this->info('Draft file already exists');
-
-            return;
         }
 
-        $stub = $this->files->stub('draft.stub');
-
-        $this->files->put('draft.yaml', $stub);
+        if(!$this->files->exists('draft.yaml')) {
+            $stub = $this->files->stub('draft.stub');
     
-        $this->info('Created example draft.yaml file in project root');
+            $this->files->put('draft.yaml', $stub);
+        
+            $this->info('Created example draft.yaml file in project root');
+        }
 
         $this->call('blueprint:trace');
     }
--- a/src/Commands/StartCommand.php
+++ b/src/Commands/StartCommand.php
@@ -43,7 +43,7 @@ class StartCommand extends Command
     public function handle()
     {
         if ($this->files->exists('draft.yaml')) {
-            $this->error('Draft file already exists');
+            $this->info('Draft file already exists');
 
             return;
         }
--- a/src/Commands/StartCommand.php
+++ b/src/Commands/StartCommand.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace Blueprint\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Filesystem\Filesystem;
+
+class StartCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'blueprint:start';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Create example draft.yaml file in project root';
+
+    /** @var Filesystem $files */
+    protected $files;
+
+    /**
+     * @param Filesystem $files
+     * @param \Illuminate\Contracts\View\Factory $view
+     */
+    public function __construct(Filesystem $files)
+    {
+        parent::__construct();
+
+        $this->files = $files;
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        if (file_exists($this->defaultDraftFile())) {
+            $this->error('Draft file already exists');
+
+            return;
+        }
+
+        $stub = $this->files->stub('draft.stub');
+
+        $this->files->put('draft.yaml', $stub);
+    
+        $this->info('Created example draft.yaml file in project root');
+    }
+
+    private function defaultDraftFile()
+    {
+        if (file_exists('draft.yaml')) {
+            return 'draft.yaml';
+        }
+
+        if (file_exists('draft.yml')) {
+            return 'draft.yml';
+        }
+
+        return null;
+    }
+}
--- a/src/Commands/StartCommand.php
+++ b/src/Commands/StartCommand.php
@@ -54,6 +54,6 @@ class StartCommand extends Command
     
         $this->info('Created example draft.yaml file in project root');
 
-        // run blueprint:trace
+        (new TraceCommand(new Filesystem))->handle();
     }
 }
--- a/src/Commands/StartCommand.php
+++ b/src/Commands/StartCommand.php
@@ -42,7 +42,7 @@ class StartCommand extends Command
      */
     public function handle()
     {
-        if (file_exists($this->defaultDraftFile())) {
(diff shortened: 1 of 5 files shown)


ProjektGopher:
Not sure how to fake the existence to draft.yaml for first test, Or how to properly call the handle method on the start command in tests. Aside from that, this command just copies draft.stub into the project root.

jasonmccreary:
Maybe you should team up with @Wowenko. You both just opened a WIP PR for the same thing...

src/Commands/StartCommand.php
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        if ($this->files->exists($this->defaultDraftFile())) {
+            $this->error('Draft file already exists');
```
jasonmccreary, line 46: No need to error. We can gracefully play on.

src/Commands/StartCommand.php
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        if ($this->files->exists($this->defaultDraftFile())) {
```
jasonmccreary, line 45: Just use `draft.yaml` for this command.

src/Commands/StartCommand.php
```diff
@@ -0,0 +1,70 @@
...
+            $this->error('Draft file already exists');
+
+            return;
+        }
+
+        $stub = $this->files->stub('draft.stub');
+
+        $this->files->put('draft.yaml', $stub);
+    
+        $this->info('Created example draft.yaml file in project root');
```
jasonmccreary, line 55: I'd like to also run the `trace` command as part of this.

stubs/draft.stub
```diff
@@ -0,0 +1,20 @@
+models:
```
jasonmccreary, line 1: At the most, the stub should make a `models` and `controllers` block. Maybe only this `# ...` underneath.

Otherwise, an empty file is fine.

tests/Feature/StartCommandTest.php
```diff
@@ -0,0 +1,38 @@
...
+use Blueprint\Commands\StartCommand;
+use Illuminate\Filesystem\Filesystem;
+use Tests\TestCase;
+
+class StartCommandTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function it_exits_with_error_if_draft_file_already_exists()
```
jasonmccreary, line 15: This can be dropped.

tests/Feature/StartCommandTest.php
```diff
@@ -0,0 +1,38 @@
...
+
+        // how to check that it exitted with error?
+    }
+
+    /**
+     * @test
+     */
+    public function it_creates_a_draft_file_if_none_exists()
+    {
+        $file = \Mockery::mock(Filesystem::class);
```
jasonmccreary, line 32: You can use `swap` here to mock + switch out `Filesystem` in the container.

tests/Feature/StartCommandTest.php
```diff
@@ -0,0 +1,38 @@
...
+        $file->expects('put')
+        ->with('draft.yaml');
+
+        (new StartCommand(new Filesystem))->handle();
```
jasonmccreary, line 36: Then pass in `file` here.

Wowenko:
Hey @ProjektGopher I am not at home but it seems you are close to achieving it, cheer up!

tests/Feature/StartCommandTest.php
ProjektGopher: Took me a while, but I did find that one, haha.

stubs/draft.stub
ProjektGopher: Truncated draft.stub

tests/Feature/StartCommandTest.php
ProjektGopher: Removed

src/Commands/StartCommand.php
ProjektGopher: Should I just call TraceCommand(FileSystem::class)->handle() directly, or is there another better way I should be doing that?

ProjektGopher:
My test always says "blueprint:start command does not exist" and I cannot figure out why... It also says the same thing with every other blueprint command. I tried adding resolve(StartCommand::class) to add it before calling it in the test, but the app still can't see the commands. When I replace blueprint:* with other artisan commands, like 'list' or 'help' I don't get that error.

Pr3d4dor:
@ProjektGopher I thin you need to add this method to base TestClass:
```
protected function getPackageProviders($app)
    {
        return [
            BlueprintServiceProvider::class,
        ];
    }
```

jasonmccreary:
Agreed. This is probably a [TestBench](https://github.com/orchestral/testbench) thing due to the _deferred_ service provider.

ProjektGopher:
The tests seem to be passing now!

The one last thing I have to figure out is whether trace only runs when we create a new draft file, or if it just run after blueprint:start regardless

jasonmccreary:
This command should create `draft.yaml` (if it doesn't exist) and run `blueprint:trace`.

ProjektGopher:
The tests are passing now, but for whatever reason when I run the tests, it actually creates a draft.yaml file in the root.

I also seem to be able to *anything* in shouldReceive() and it still passes... not sure what the deal is

jasonmccreary:
Yeah, this a gotcha with first time testers. That method _should_ really be named something like `mayReceive` 😉

No worries. I'm probably going to sit on this until Thursday and clean it up then. Of course, you're welcome to work through it if you're on a mission.

Otherwise, this is great. I'll get it merge soon.

ProjektGopher:
I'm using this as a learning experience, so I'll keep working on it tomorrow. If I don't figure it out I'll definitely be paying close attention when you fix it on Thursday's stream.

I've added \Mockery::close() at the end of each of my tests but now none of the assertions are being run, and it's still hitting the real Filesystem::put() method, creating a draft.yaml every time I run the tests.

Didn't expect testing such a simple feature to be this agonizing.

tests/Feature/StartCommandTest.php
```diff
@@ -0,0 +1,56 @@
...
+    public function it_does_not_create_a_draft_file_if_one_exists_already()
+    {
+        Filesystem::shouldReceive('exists')
+            ->andReturn(true);
+
+        Filesystem::shouldNotReceive('put');
+
+        Artisan::call('blueprint:start');
+
+        \Mockery::close();
```
jasonmccreary, line 44: It's interesting this is required with TestBench as Laravel includes this... I'll take a peak and potentially send them a PR.

jasonmccreary:
Thanks for all your work on this. There's a few things I want to change. So I'll tidy it up tomorrow.

ProjektGopher:
Sounds good, man. I'll try to put a little bit more time into it this evening if I'm able.

I've fixed a spacing issue locally that had one of your checks squawking, I've got a an idea about why it's running the real put method on the filesystem - so I'll see what I can do about that, I'll try to replace shouldReceive with some proper assertions, and I'll try to finish the test checking that blueprint:trace is always run.

Didn't expect such a small feature to take me so much time to implement, but it's been a great way for me to learn.

jasonmccreary:
Since I can't push directly to your branch (thanks GitHub), I had to close this and [open a new one](#192).

However, for future readers. The issue with the failing tests were:

- Although these are bound in the container, we were mocking the `Filesystem` facade, not the actual Filesystem "contact" which was injected to the constructor.
- To mock this "contract", you have to [bind it to "files"](https://laravel.com/docs/7.x/facades#facade-class-reference). This was the [issue during the live-stream](https://www.youtube.com/watch?v=GnTN-mKvqqo&feature=youtu.be&list=PLmwAMIdrAmK5rH3SWvokHV8xI_0mauxDL&t=6412).
- Since this command in turn calls the `trace` command, which does has its own behavior with the filesystem, you have to make it a _partial_ mock. Otherwise, you'd need to mock all it's calls. I wish this was easier to test in Laravel.

PR #202, opened 2020-05-14
Added Polymorphic Relationship

RahulDey12:
Fix against #199
New Syntax Should Be Like

```
Post:
    title: string:400
    published_at: timestamp nullable
    relationships:
      morphMany: Comment

  Comment:
    title: string
    relationships:
      morphTo: null
```

PR #204, opened 2020-05-15
feat: support for polymorphic relations

### LINKED ISSUE #199: Polymorphic Relations are not supported
### Synopsis:
Polymorphic Relations are not supported

### Proposed Syntax:
`morphTo: attendable`
`morphMany: Attendance`

### Expected Behavior:

That should add a Polymorphic relation [](https://laravel.com/docs/7.x/eloquent-relationships#many-to-many-polymorphic-relations)

Pr3d4dor:
This pull request adds support for polymorphic relations.

This approach can be optimized.

The problem is that we need to add the `morphTo` method to the related model on the `morphOne` and `morphMany`.

Example, given the following definition:

```
models:
  Post:
    name: string:400
    relationships:
      morphMany: Image

  User:
    name: string:400
    relationships:
      morphMany: Image

  Image:
    url: string:400
    imageable_id: integer
    imageable_type: string:400
```

We need to add `morphMany` to `Post` and `User` models, but we also need to add the `morphTo` method to the `Image` model. Using this approach we solve the need to add `morphTo: null` to the definition, like the #202 pull request.

I think that, adding the `morphTo` relationship before the output, to the referenced model, is a "cheat". But I could not come up with a cleaner way to do this.

---
Supercedes #202
Closes #199


jasonmccreary:
Thanks. This makes much more sense now.

In seeing the draft, I propose the following syntax:

```yaml
models:
  Post:
    name: string:400
    relationships:
      morphMany: Image

  User:
    name: string:400
    relationships:
      morphMany: Image

  Image:
    url: string:400
    relationships:
      morphTo: Imageable
```

This would overcome the `morphTo: null` issue, yet give Blueprint the information it needs to properly build the relationship. It also prevents the developer from having to remember the format/type of the morph columns.

src/Generators/ModelGenerator.php
```diff
@@ -24,8 +26,12 @@ public function output(array $tree): array
 
         $stub = $this->files->stub('model/class.stub');
 
+        $this->registerModels($tree);
+
+        $this->mapPolymorphicRelationships();
+
         /** @var \Blueprint\Models\Model $model */
-        foreach ($tree['models'] as $model) {
+        foreach ($this->models as $model) {
```
jasonmccreary, line 34: You don't want to loop over every model. This will regenerate "cached" models as well.

tests/fixtures/models/image-polymorphic-relationship.php
```diff
@@ -0,0 +1,34 @@
...
+class Image extends Model
+{
+    /**
+     * The attributes that are mass assignable.
+     *
+     * @var array
+     */
+    protected $fillable = [
+        'url',
+        'imageable_id',
```
jasonmccreary, line 16: Probably should not be fillable. However, I think the new syntax would solve this.

Pr3d4dor:
@jasonmccreary Thanks, tonight I will refactor this pull request.

Pr3d4dor:
@jasonmccreary I've updated the pull request.

There is one thing that is missing thought, the generate migration for the `Image` model does not include the `imageable_id` and `imageable_type` columns. Do I need to implement this behavior in the `MigrationsGenerator`?

jasonmccreary:
@Pr3d4dor, yes. I would think for whatever models have the `morphTo` relationships, to use the provided name to create two columns, one for the `id` and one for the `type`. Ideally these are right before the `timestamps`.

Pr3d4dor:
@jasonmccreary I've updated the pull request adding the support for polymorphic relations in migrations.

There is one adjustment that needs to be made: The migration for the table that contains `morphTo` needs to be before the other ones.

jasonmccreary:
Looks pretty good. I have some small changes I'll make.

Why does it need to come first?

Pr3d4dor:
Sorry :joy:, I confused myself looking in the Laravel Documentation. It doesn't need to come first.

mahfoudfx:
> Thanks. This makes much more sense now.
> 
> In seeing the draft, I propose the following syntax:
> 
> ```yaml
> models:
>   Post:
>     name: string:400
>     relationships:
>       morphMany: Image
> 
>   User:
>     name: string:400
>     relationships:
>       morphMany: Image
> 
>   Image:
>     url: string:400
>     relationships:
>       morphTo: Imageable
> ```
> 
> This would overcome the `morphTo: null` issue, yet give Blueprint the information it needs to properly build the relationship. It also prevents the developer from having to remember the format/type of the morph columns.

Hello, I think that it will be very helpful if you could update the official [documentation](https://blueprint.laravelshift.com/docs/model-relationships/) to include polymorphic relationships syntax.

jasonmccreary:
@mahfoudfx, totally agree. A bit busy at the moment. Would you be willing to [submit a PR to the docs repo](https://github.com/laravel-shift/blueprint-docs) to help get this started?

mahfoudfx:
@jasonmccreary , I've submitted a PR. I hope it's correct or require a little changes. Also, I tried also to write something regarding morphToMany and hasManyThrough, but I found nothing and I'm not sure if blueprint handle them.

PR #221, opened 2020-05-20
feat: ability to customize pivot table columns (belongsToMany)

### LINKED ISSUE #175: Add ability to customize pivot columns
Hi, amazing work by the way!

### Current situation

Reviewing the docs I can see that you can specify `belongsToMany` relationships like so:

```yaml
# draft.yaml

models:
  User:
    ...
    relationships:
      belongsToMany: Store
```

This will generate something similar to the following method in the `User` model:

```php
# User.php

public function stores()
{
    return $this->belongsToMany(Store::class);
}
```
But, currently, there's no way to customize the pivot table name: `store_user`

### Proposal

It could be possible to pass an attribute when specifying the relationship to customize the pivot table:

```yaml
models:
  User:
    ...
    relationships:
      belongsToMany: Store:favorites
```

Generating:

```php
# User.php

public function stores()
{
    return $this->belongsToMany(Store::class, 'favorites');
}
```

**Even further**, in case you specify the pivot colum as a PascalCase word, this could generate the pivot model and configure it in the related models:

```yaml
#draft.yaml

models:
  User:
    ...
    relationships:
      belongsToMany: Store:Favorite
```

```php
# User.php

public function stores()
{
    return $this->belongsToMany(Store::class)->using(Favorite::class);
}
```

Pr3d4dor:
This pull request adds the ability to customize pivot table columns (belongsToMany).

---
Closes #175

### DIFF
--- a/src/Generators/ModelGenerator.php
+++ b/src/Generators/ModelGenerator.php
@@ -137,29 +137,46 @@ class ModelGenerator implements Generator
 
         foreach ($model->relationships() as $type => $references) {
             foreach ($references as $reference) {
+                $key = null;
+                $class = null;
+
+                $column_name = $reference;
+                $method_name = Str::beforeLast($reference, '_id');
+
                 if (Str::contains($reference, ':')) {
-                    [$class, $name] = explode(':', $reference);
-                } else {
-                    $name = $reference;
-                    $class = null;
+                    [$foreign_reference, $column_name] = explode(':', $reference);
+                    $method_name = Str::beforeLast($column_name, '_id');
+
+                    if (Str::contains($foreign_reference, '.')) {
+                        [$class, $key] = explode('.', $foreign_reference);
+
+                        if ($key === 'id') {
+                            $key = null;
+                        } else {
+                            $method_name = Str::lower($class);
+                        }
+                    } else {
+                        $class = $foreign_reference;
+                    }
                 }
 
-                $name = Str::beforeLast($name, '_id');
-                $class = Str::studly($class ?? $name);
+                $class = Str::studly($class ?? $method_name);
 
                 if ($type === 'morphTo') {
                     $relationship = sprintf('$this->%s()', $type);
                 } elseif ($type === 'morphMany' || $type === 'morphOne') {
-                    $relation = Str::of($name)->lower()->singular() . 'able';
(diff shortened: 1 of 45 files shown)


src/Generators/ModelGenerator.php
```diff
@@ -145,9 +145,18 @@ private function buildRelationships(Model $model)
...
                 $name = Str::beforeLast($name, '_id');
-                $class = Str::studly($class ?? $name);
-                $relationship = sprintf("\$this->%s(%s::class)", $type, '\\' . $model->fullyQualifiedNamespace() . '\\' . $class);
-
+                $relationShipClass = Str::studly($class ?? $name);
+
+                $relationship = "";
+                if (!is_null($class) && $type === 'belongsToMany') {
+                    if (preg_match("/^[A-Z]/", $name)) {
+                        $relationship = sprintf("\$this->%s(%s::class)->using(%s::class)", $type, '\\' . $model->fullyQualifiedNamespace() . '\\' . $relationShipClass, '\\' . $model->fullyQualifiedNamespace() . '\\' . $name);
```
jasonmccreary, line 153: Please use single quote for strings.

src/Generators/ModelGenerator.php
```diff
@@ -145,9 +145,18 @@ private function buildRelationships(Model $model)
...
 
                 $name = Str::beforeLast($name, '_id');
-                $class = Str::studly($class ?? $name);
-                $relationship = sprintf("\$this->%s(%s::class)", $type, '\\' . $model->fullyQualifiedNamespace() . '\\' . $class);
-
+                $relationShipClass = Str::studly($class ?? $name);
+
+                $relationship = "";
+                if (!is_null($class) && $type === 'belongsToMany') {
+                    if (preg_match("/^[A-Z]/", $name)) {
```
jasonmccreary, line 152: It's not clear that this `preg_match` determines if the developer passed a _model reference_.

I am open to supporting this. But I feel there are more implications (e.g. generating the pivot model). For now, I think we should focus on the original issue, which was specifying the custom table name.

Said another way, let's assume, for now, that whatever they pass is a custom table name.

Pr3d4dor:
Ok, tomorrow I will update the pull request with the requested changes.

jasonmccreary:
@Pr3d4dor, no rush. Thanks.

PR #230, opened 2020-05-22
feat: add support for `cascade`, `set null`, `restrict`, and `no action` for onDelete actions

### LINKED ISSUE #213: Allow to specify the restriction type on a foreign key
### Synopsis:
According the documentation, Blueprint 'will generate code to add the reference and cascade "on delete" constraints.' It would be very handy if the type of restriction could be specified being cascade, restrict, set_null, no_action or set_default.

### Proposed Syntax:
Allow the third parameter after foreign table to describe the restriction.

`  foreign_table_id: foreign_table:<restriction type> nullable`

### Expected Behavior:
Definition:
`domain_id: id:domain:set_null nullable`

Should generate the following code:
`$table->foreignId('domain_id')->nullable()->constrained()->onDelete('SET NULL');`

In particular I want to point out that the 'nullable()' instruction should come before 'constrainted()' which is currently not the case.

nathane:
- Update `ModelLexer` to parse `onDelete` modifier
- Update `MigrationGenerator` to generate  both `cascade` and `set_null` clauses.
- Update fixtures to Test `MigrationGenerator`.

Usage:
```
models:
  Cart:
    name: string
    user_id: id foreign nullable ondelete:set_null
```
Closes #213

### DIFF
--- a/config/blueprint.php
+++ b/config/blueprint.php
@@ -94,4 +94,24 @@ return [
     */
 
     'use_guarded' => false,
+
+    /*
+    |--------------------------------------------------------------------------
+    | Foreign Key ON DELETE action
+    |--------------------------------------------------------------------------
+    |
+    | By default, Blueprint will set the `ON DELETE` action to 'cascade'.
+    |
+    | restrict / no action:
+    |   - No action is performed with the child data when the parent data is deleted.
+    | cascade:
+    |   - The child data is either deleted when the parent data is deleted.
+    | set null:
+    |   - The child data is set to NULL when the parent data is deleted.
+    |
+    | Supported: 'cascade', 'set_null', 'restrict', 'no_action'
+    |
+    */
+
+    'on_delete' => 'cascade',
 ];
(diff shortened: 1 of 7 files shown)


jasonmccreary:
If we're going to do this, we need to cover all bases:

1. Add a config option to make it easier to set the cascade option.
2. Ensure we are supporting all options: `delete`, `set null`, etc...

nathane:
- Add a config option for Foreign Key `ON DELETE` actions
- Add support for `cascade`, `set_null`, `restrict`, and `no_action` onDelete actions

> ######  RESTRICT / NO ACTION
> no action is performed with the child data when the parent data is deleted.
> ######  CASCADE
> the child data is either deleted when the parent data is deleted.
> ######  SET NULL
> the child data is set to NULL when the parent data is deleted.

jasonmccreary:
Thanks for being so thorough. I'm going to go ahead and merge this, but probably add a test for the `ondelete` syntax as well as tweak the config comment.

PR #233, opened 2020-05-23
Add blueprint:rebuild command

stoll:


jasonmccreary:
Thanks for this, but this code likely won't run in its current state.

Furthermore, if I were to add this command, I'd want to attempt to modify it's output to be a bit friendlier as far as what it regenerated and what might need adjustment...

PR #235, opened 2020-05-23
feat: add ability to differentiate between mail and notification

### LINKED ISSUE #20: Notification Vs Mail
I see a stament 

`send: ReviewNotification to post.author`

Which means that it need to Create a Notification class

`send: ReviewMail to post.author`

Which means that it need to Create a Mail class

Pr3d4dor:
This pull request adds the ability to differentiate between mail and notification when using `send`.

I created another generator called `NotificationGenerator` that basically is equal to the `MailGenerator`. The only difference is the stub that is used.

When using send, it will look for the `Notification` prefix, if the prefix is not present it will assume that is a `Mail` and not a `Notification`.

It uses the `Notification` facade, but it does not add the `Notifiable` trait to the referenced model. 

---
Closes #20


jasonmccreary:
This is good. I do wonder if we should dedicate its own syntax though, like `notify`, `broadcast`, or `push`. That way the _Notification_ suffix isn't required...

Pr3d4dor:
Yeah, I think with it's own syntax the code will be cleaner and more explicit. If you give me a `draft.yml` file I can update this pull request. Or if you prefer to tweak this pull request in the live stream I can wait and watch the live stream later (I can't watch live because of my work schedule).

jasonmccreary:
After [reading the docs](https://laravel.com/docs/7.x/notifications#using-the-notifiable-trait), I wonder if there is room for both.

```yaml
notify: user NewInvoice:invoice, data, whatever

send: NewInvoiceNotification to:users with:invoice
```

```php
$user->notify(new InvoicePaid($invoice));

Notification::send($users, new InvoicePaid($invoice));
```

Basically you already have the second example now. We'd just add the first. What do you think?

Pr3d4dor:
I liked it, tomorrow if I have time, I will implement the other one.

The `Notifiable` trait needs to be added automatically to the related model when using `notify` or either one?

jasonmccreary:
I would say on the new one, ideally we should add the `Notifiable` trait to any model we are generating in the current draft.

We haven't crossed the bridge of updating existing models yet. So users will just have to add it otherwise.

If you get to it, great. Otherwise, I'll merge this add add the `notify` in the next live-stream.

jasonmccreary:
@Pr3d4dor is this done?

Pr3d4dor:
@jasonmccreary Not yet, it doesn't support the syntax that you proposed. I will work on it this week, because last week I was very busy.

jasonmccreary:
No problem at all.

Just trying to understand if this first portion is in a mergeable state so you don't have to keep battling conflicts.

If I can merge this and you (or I) add the other piece later...

Pr3d4dor:
This portion is mergeable, you can merge this and a will make another pull request supporting the another syntax. Or you can wait and I will update this pull request and you can merge after that.

jasonmccreary:
Okay. I'll merge this as not to let it get stale. However, I'm not going to tag Blueprint until the other syntax is added.

PR #242, opened 2020-05-24
Update ControllerLexer.php

dishereandy:
Proposed fix for checking if resource method is prefixed with API. Solution provided by @nathane .

### DIFF
--- a/src/Lexers/ControllerLexer.php
+++ b/src/Lexers/ControllerLexer.php
@@ -34,7 +34,7 @@ class ControllerLexer implements Lexer
                 
                 // Checks if resource is prefixed with API
 
-                if (Str::startsWith(Str::lower($definition['resource']),'api') {
+                if (Str::startsWith(Str::lower($definition['resource']),'api')) {
                     $controller->setApiResource(true);
                 }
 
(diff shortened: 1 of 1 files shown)


jasonmccreary:
This doesn't necessary solve the use case of mixed actions. For example, `api.show, web.store`.

Ideally, this would also ensure there are no `web.` prefixes...

dishereandy:
Yeah, I was too focused on api that I forgot about web.

dishereandy:
From what I can gather, I think a new function for checking the prefixes of each of the controller's methods is needed. I'm not sure but I think right now, it is only checking once for whether the resource is api or web for each controller and sets the `setApiResource` to true which negates everything written behind the first method. Currently if I use `resource: web.index, api.store`, `setApiResource` will be set to false for every method and likewise if I use `resource: api.store, web.index`, it will be set to true for every method.

jasonmccreary:
I agree.

As I've said, ideally this would only consider it an _API resource_ if it were `api` and none of the actions started with `web.`.

You have half of that logic. Just write the other half ;)

dishereandy:
Haha, I know the logic but I don't how to write the code. I'll try though but as I have some projects to do, I don't think I can dedicate a lot of time to this. In the mean time I'll leave the issue and the pull request,  in case somebody  figures out a solution.

PR #244, opened 2020-05-25
feat: streamline draft definitions for fixtures

nathane:
- Change all fixtures definitions file extensions from `*.bp`to `*-draft.yaml`.


jasonmccreary:
To streamline this just a bit, let's have the extension as simply `.yaml` and instead change the folder name from `definitions` to `drafts`.

PR #249, opened 2020-05-28
Build command improvements

diogogomeswww:
2 Changes:

- If the draft file is not found it will exit the command, preventing exceptions.
Before:
image
After:
image

- Add info for the `draft argument

Right now there's no info about the optional argument:
before

With this PR we will have the info for the `draft` argument:
after


jasonmccreary:
Nice little paper cut fix. Thanks.

PR #253, opened 2020-05-30
feat: support notification via model

Pr3d4dor:
This pull request adds the support to use `Notifications` using the model `notify` method.

The current implementation does not add the `Notifiable` trait to the referenced model.


jasonmccreary:
@Pr3d4dor kind of forgot about this. What's the current state?

Pr3d4dor:
The PR is working and already fixes the #274 issue, you can merge it if you want.

The only thing missing is:

> The current implementation does not add the Notifiable trait to the referenced model.

jasonmccreary:
Gotcha. Thanks.

PR #276, opened 2020-06-26
Fix #247: Implements blueprint:build --only and --skip

### LINKED ISSUE #247: Allow blueprint:build to generate only a subset of things (e.g. factories, tests, etc.)
### Synopsis:
When using Blueprint in an existing Laravel project, you might want to generate 'only' a subset of things. For example: Say you have a code base that does not have factories, then you might want to generate 'just' the factories.

This might also be very handy for 'updating' the Model's `$fillables` once the definition changes in the `draft.yaml`.

### Proposed Syntax:
```sh
php artisan blueprint:build --only=factories
```

Where `--only` takes comma separated list of 'things' to generate.

Those `things` being:
* controllers
* factories
* migrations
* models
* routes
* seeders
* tests

Another syntax could be:
```sh
php artisan blueprint:build [draft] [thing]
```
But that might be too 'shorthand-y'

You could also do some kind of inverse:
```sh
php artisan:build --skip=migrations
```
This would generate everything but migrations.

For instance, I would definitely use a `--skip=routes,migrations` all the time, because of the timestamps in the filenames of the migrations, and the appending of the routes in `web.php` and `api.php`.

### Expected Behavior:
```sh
> php artisan blueprint:build --only=factories
- database/factories/CertificateFactory.php
- database/factories/CertificateTypeFactory.php
```

axit-joost:
Closes #247 

This is an implementation of the `--only` and `--skip` command line options for the `blueprint:build` command.

Both commands take a comma separated list of 'file classes' (for the lack of a better word), those being:

* controllers
* factories
* migrations
* models
* routes
* seeders
* tests
* events
* requests
* jobs
* mails
* notifications
* resources
* views

The command passes the string values to `Blueprint\Builder::execute()` as additional, optional parameters `$only` and `$skip`. The action body will convert these strings to arrays, and passes those arrays to `$blueprint->generate()` alongside of the `$registry`. This action iterates over the registered  `$this->generators` and for each one call `output()`, again passing the `$tree`, `$only` and `$skip`.

Each registered generator implements `Blueprint\Contracts\Generator`. The contract is updated to include `array $only, array $skip` in the `output()` signature, which demands all of the generators adhere to this.

Inside the `output()` action of each generator, I have wrapped the output generation inside an `if` statement that references a `$this->shouldGenerate()` method, passing the `$only` and `$skip` to that method, which returns a boolean value whether it should generate or not.

Currently, the `shouldGenerate()` utilizes `in_array()` calls, using a hardcoded string to identify the 'file class'. Perhaps this should be extracted to a protected instance variable, but I am unsure what to call it.

With regards to the test suite: of course this blew up everything, as everything Mocked needed to be modified. In this Work in Progress I've managed to get the test suite to green again, but I still need to implement test cases to verify the proper workings of both command line options.

Now, before I do that, I thought it might be wise to put it under early review before continuing.

I am looking forward to all your feedback, so that I can modify/proceed accordingly.

### DIFF
--- a/src/Blueprint.php
+++ b/src/Blueprint.php
@@ -38,19 +38,19 @@ class Blueprint
         }
 
         $content = preg_replace_callback('/^(\s+)(id|timestamps(Tz)?|softDeletes(Tz)?)$/mi', function ($matches) {
-            return $matches[1].strtolower($matches[2]).': '.$matches[2];
+            return $matches[1] . strtolower($matches[2]) . ': ' . $matches[2];
         }, $content);
 
         $content = preg_replace_callback('/^(\s+)(id|timestamps(Tz)?|softDeletes(Tz)?): true$/mi', function ($matches) {
-            return $matches[1].strtolower($matches[2]).': '.$matches[2];
+            return $matches[1] . strtolower($matches[2]) . ': ' . $matches[2];
         }, $content);
 
         $content = preg_replace_callback('/^(\s+)resource?$/mi', function ($matches) {
-            return $matches[1].'resource: web';
+            return $matches[1] . 'resource: web';
         }, $content);
 
         $content = preg_replace_callback('/^(\s+)uuid(: true)?$/mi', function ($matches) {
-            return $matches[1].'id: uuid primary';
+            return $matches[1] . 'id: uuid primary';
         }, $content);
 
         return Yaml::parse($content);
@@ -70,12 +70,12 @@ class Blueprint
         return $registry;
     }
 
-    public function generate(array $tree): array
+    public function generate(array $tree, array $only = [], array $skip = []): array
     {
         $components = [];
 
         foreach ($this->generators as $generator) {
-            $components = array_merge_recursive($components, $generator->output($tree));
+            $components = array_merge_recursive($components, $generator->output($tree, $only, $skip));
         }
 
         return $components;
--- a/src/Blueprint.php
+++ b/src/Blueprint.php
@@ -14,7 +14,7 @@ class Blueprint
 
     public static function relativeNamespace(string $fullyQualifiedClassName)
     {
-        $namespace = config('blueprint.namespace') . '\\';
+        $namespace = config('blueprint.namespace').'\\';
         $reference = ltrim($fullyQualifiedClassName, '\\');
 
         if (Str::startsWith($reference, $namespace)) {
@@ -38,19 +38,19 @@ class Blueprint
         }
 
         $content = preg_replace_callback('/^(\s+)(id|timestamps(Tz)?|softDeletes(Tz)?)$/mi', function ($matches) {
-            return $matches[1] . strtolower($matches[2]) . ': ' . $matches[2];
+            return $matches[1].strtolower($matches[2]).': '.$matches[2];
         }, $content);
 
         $content = preg_replace_callback('/^(\s+)(id|timestamps(Tz)?|softDeletes(Tz)?): true$/mi', function ($matches) {
-            return $matches[1] . strtolower($matches[2]) . ': ' . $matches[2];
+            return $matches[1].strtolower($matches[2]).': '.$matches[2];
         }, $content);
 
         $content = preg_replace_callback('/^(\s+)resource?$/mi', function ($matches) {
-            return $matches[1] . 'resource: web';
+            return $matches[1].'resource: web';
         }, $content);
 
         $content = preg_replace_callback('/^(\s+)uuid(: true)?$/mi', function ($matches) {
-            return $matches[1] . 'id: uuid primary';
+            return $matches[1].'id: uuid primary';
         }, $content);
 
         return Yaml::parse($content);
@@ -75,7 +75,9 @@ class Blueprint
         $components = [];
 
         foreach ($this->generators as $generator) {
-            $components = array_merge_recursive($components, $generator->output($tree, $only, $skip));
+            if ($this->shouldGenerate($generator->types(), $only, $skip)) {
+                $components = array_merge_recursive($components, $generator->output($tree));
+            }
         }
 
         return $components;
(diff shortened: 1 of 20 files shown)


dmason30:
Seems like a nice addition 🚀 

However, I am not sure adding the same `shouldGenerate` logic to every generator is the correct approach? Why not just have a method that returns the type e.g.

`Blueprint\Contracts\Generator` interface;
```php
public function type(): string;
```

In each of the generators e.g. Blueprint\Generators\TestGenerator;
```php
public function type() 
{
     return 'tests';
}
```

Then the `shouldGenerate` logic should be handled in the `Blueprint::generate` method and getting the type for each generator from the above.

I would also suggest moving the type strings into constants somewhere maybe its own class e.g. `\Blueprint\GeneratorTypes`

axit-joost:
Thank you @dmason30, this was exactly the kind of feedback I was looking for. I'll take a stab at it probably somewhere this week, and also do the testing.

I have never tested command line utilities before, so I'm still looking at the best way to go about this. Do you have any pointers for me in that regard? Perhaps @jasonmccreary could also chip in on what would be the right way to go about this?

Thanks all!

axit-joost:
@dmason30 I took your suggestion and applied it. The `Builder` class now houses a modified version of `shouldGenerate` that takes a `$generator->types()` array of strings representing the types of files generated, and returns a boolean that is used by the `if` condition to determine whether or not it should call the output on the generator.

That is indeed a much nicer, and more DRY approach, than having that `shouldGenerate()` in each and every generator.

I have also added, as per your suggestion, the `array` of `strings` approach and added `['controllers', tests']` to the TestGenerator. Although I am not entirely convinced of its usefulness. I do get what you are saying, and I do believe that there might be use cases that have use for this, so: it's there.

The only thing missing is the tests. As mentioned earlier, I am not sure how to approach this in the test code. I'm looking at `StartCommandTest` that uses a Mock filesystem, but that is more like a Unit test. I'm looking more at a Feature test that runs the `blueprint:build` command with a specific `draft.yaml` from the fixtures and specific options, to be able to assert whether or not files are being generated and output.

If anyone can make a suggestion, that would be great.

axit-joost:
When testing by hand on a blank project, using the `pascal-case` fixture as `draft.yaml` file, I stumbled upon the fact that the `FormRequestGenerator` has logic inside of it that checks if a Request output file already exists and skips it if it does. My question would be: why is this code there, and for what purpose?

jasonmccreary:
@axit-joost, I think certain _resources_ make sense to skip, not overwrite. For example, form requests, a view, jobs, mailables, etc.

Most of Blueprint is references and assumptions based on convention. So if one of these _references_/_resources_ already exists, I wouldn't want to overwrite it.

As such, I do consider the above different from what you're proposing. So I do feel having every generator be responsible for knowing if it should be skipped or generated seems like a harder approach than Blueprint simply not calling that generator.

To make it easier, I wouldn't expect you to be able to skip _low-level_ components. My expectation is you could skip things like controllers, models, views, tests, factories. Not mailables, jobs, etc. Otherwise Blueprint would generate a lot of incomplete/invalid code.

axit-joost:
@jasonmccreary thanks for your feedback.

Gathering from your response, I compiled this list:

| Generator | Allow --only or --skip | Auto-skips not to overwrite |
|---|:---:|:---:|
| Controller | X |
| Factory | X |
| Migration | X |
| Model | X |
| Route | X |
| Seeder | X |
| Test | X |
| Event |  | X |
| FormRequest | X | X |
| Job |  | X |
| Mail  |  | X |
| Notification  |  | X |
| Resource | X | X |
| View | X | X |

To verify: The above `X`-marked Generators SHOULD  be responsible for knowing if it should be skipped or generated.

The downside, in my opinion, is that you will have a `shouldGenerate()` method in all of those Generators, which is not very DRY. However, I *do* concur with your stance of not being able to skip over low level stuff like Jobs and Mailables, because it would basically amount to incomplete and invalid code.

However, for a best of both worlds, we could alternatively change the output of the `$generator->types()` method on the generators that do not have the `X` above to just be ['controllers'], because they are all statements from the `controllers:` section of the draft. 

If you then would `--skip=controllers` it would skip the controllers and the statements. If you would do `--only=controllers`, it would generate the controllers AND all of the lower level stuff that the controllers would require, therefore generating code that is complete and valid.

(Sorry for being this lengthy again, I just want to be thorough)

jasonmccreary:
I don't really know what you mean about `$generator-types()`. However, your description of how to proceed seems we're in alignment.

All I will say is I always look for the minimal change. So while I appreciate the contribution, I will not merge something that overly complicates the generators or changes the current output.

I see these options as disabling generation of certain resources. We can trust that a developer that specifies these options _knows_ the code which is or is not generated.

axit-joost:
I think this is done.

I believe this hits the sweet spot of:

* **Being a minimal change** that does not overly complicates the generators or changes the current output: the only addition is a `public function types(): array` which returns an `array` of `strings` that denote the types of files generated. This concern is added to the `Generator` `Contract`.

* **The code is DRY**: Instead of repeating the same logic in every generator, we have a single `shouldGenerate()` on the `Blueprint` class, which is called in the `generate()` method. This also keeps the changes to the actual generators minimal.

* **The code does not change the output contents**: as a matter of fact, if you use Blueprint same as before, nothing changes. It does what it says on the tin and skips and limits accordingly.

* **The feature exempts low-level components**: The user is unable to address Events, Mailables, Jobs and Notifications to be skipped or limited individually. Instead, it can be skipped or limited using `--skip=controllers` or `--only=controllers` for wholesale skipping/only generation.

* **The code is tested**: The test suite is extended to test both the --only as well as the --skip option for both single as well as multiple values (comma separated).

The only thing is that my local php-cs-fixer may have done something to the preferred whitespace (e.g. concat spacing, et al.). So you might want to hit 'ignore whitespace' when reviewing the code.

If you want me to fix this, or any other issues, please let me know.

jasonmccreary:
Cool. Merged. I'll let it marinate on `master` for a few days and then tag/announce it. Feel free to point your dependency at `dev-master` to test it out more. 👍

adnoh:
it's a nice feature. would be great if you can extend the documentation and the information on the command help text by the list of types one can use. If you aren't used to the internal types of blueprint - like me - you'r lost and need to dig into git to find it.
What I would have use for is to regenerate only for example one specific controller - so I would like to give the controllername or filename to just regenerate this one

bhanwarpsrathore:
@jasonmccreary This is nice feature to be used, but its not there in documentation. Updating those will be useful. Thanks in advance

PR #281, opened 2020-07-04
Add reference factory support to validation shorthand

dmason30:
I saw you left a TODO in #280 and while i was migrating this change to my pest add-on I added support to generate the reference factories in tests. 

So with a draft of:

```yml
models:
  Certificate:
    name: string
    certificate_type_id: id:certificate_type
    reference: string
    document: string
    expiry_date: date
    remarks: text nullable

controllers:
  Certificate:
    resource: api
```

**Before**
```php
$certificate_type_id = $this->faker->randomDigitNotNull;
```

**After**
```php
$certificate_type = factory(CertificateType::class)->create();
```

**Changes**
The private `generateReferenceFactory` function has been added to the `TestGenerator` with the factory generation code recently moved there as the same logic is now applied in two locations. 

I don't really like using the array references as arguments but its the best I could come up within then given structure. 

In the addon I am maintaining I switched this  to a [PendingOutput](https://github.com/fidum/laravel-blueprint-pestphp-addon/blob/master/src/Builders/PendingOutput.php) object so that I could manage output in any function. 

I look forward to your feedback 🙌

### DIFF
--- a/src/Generators/TestGenerator.php
+++ b/src/Generators/TestGenerator.php
@@ -4,6 +4,7 @@ namespace Blueprint\Generators;
 
 use Blueprint\Blueprint;
 use Blueprint\Contracts\Generator;
+use Blueprint\Models\Column;
 use Blueprint\Models\Controller;
 use Blueprint\Models\Statements\DispatchStatement;
 use Blueprint\Models\Statements\EloquentStatement;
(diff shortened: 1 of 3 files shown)


jasonmccreary:
Thanks for this. There are a couple assumptions here that make it work. Not sure if those already existed or were changed, but I'm going to open a new PR with your work and some changes to get this all the way there.

PR #286, opened 2020-07-09
Use strict check on null instead of empty()

### LINKED ISSUE #285: Boolean default:0 returns default() instead of default(0)
- Laravel Version: 6.* and 7.*
- PHP Version: 7.4
- Blueprint Version: v1.15.0
- Platform: Linux

### Issue:

If I define a default value 0 for a boolean, in the migration it will show up as `$table->boolean('online')->default();`. Expected behavior would be: `$table->boolean('online')->default(0);`.

### draft.yaml:
```yaml
models:
  Category:
    title: string
    sort_order: integer default:10
  Page:
    title: string
    body: longtext
    category_id: id foreign:categories
    sort_order: integer default:10
    online: boolean default:0
```

jyrkidn:
This fixes #285, also added the `default:0` modifier as a test case.

### DIFF
--- a/src/Lexers/ModelLexer.php
+++ b/src/Lexers/ModelLexer.php
@@ -232,7 +232,7 @@ class ModelLexer implements Lexer
 
             if (isset(self::$modifiers[strtolower($value)])) {
                 $modifierAttributes = $parts[1] ?? null;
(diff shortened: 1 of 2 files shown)


jasonmccreary:
Nice work!

PR #297, opened 2020-07-18
Make generator classes extendable

### LINKED ISSUE #293: Extendable Classes
https://github.com/laravel-shift/blueprint/blob/f289ca7a4f46369661706cba168e8ef83c689103/src/Generators/ControllerGenerator.php#L83

It would be neat if the `buildMethods(Controller $controller)` method is accessible from inheritance, so the class can be extended, so additional methods can have their own statements for it.
The statement classes are listed within the **private** `buildMethods(Controller $controller)` method, where it checks the instance type of the statement to output the statement to the body, very cool, but it limits the possibility to add additional methods to the controller stub.  I have not been able to generate additional methods without creating my own controller generator, which i hate doing, i don't want to mis out on updates when new features are introduced to the package when updating the package.

The idea is to extend the ControllerGenerator class with a CustomControllerGenerator, swap the generator in the serviceprovider with the normal ControllerGenerator, inheriting the needed methods from de default ControllerGenerator to add additional (custom) methods with their statements.

So this will give me the ability to add my own logic, before i call the parent to wrap it up:
```php
//My logic
$parentMethods = parent::buildMethods($controller);
populate the stub after this.
```
I understand that this package is created to produce the default setup needed to start from, and not made to customise those outcomes, then again, it would be neat if possible.

Pr3d4dor:
This pull request make generator classes extendable.

---
Supercedes #296
Closes #293

### DIFF
--- a/src/Generators/ControllerGenerator.php
+++ b/src/Generators/ControllerGenerator.php
@@ -80,7 +80,7 @@ class ControllerGenerator implements Generator
         return $stub;
     }
 
-    private function buildMethods(Controller $controller)
+    protected function buildMethods(Controller $controller)
     {
         $template = $this->files->stub('controller/method.stub');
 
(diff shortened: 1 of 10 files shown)


jasonmccreary:
No worries. Thanks for ordering them as well. Nice touch. 👌

PR #299, opened 2020-07-20
Refactor shouldGenerate using collection

devmsh:
After watching some live streams about blueprint. I fount it interesting to help refactor the project a bit.

I don't know if you would be interested to receive many small patches like this one or not. so please advice.

I will not add/modify functionalities, I will only refactor the code a bit to make code cleaner and more readable for new contributions, then I may contribute by adding new features as well.

I will submit another example PR so you can give me your thoughts and direction.


jasonmccreary:
Yes, small PRs are more likely to be reviewed and merged quickly.

PR #300, opened 2020-07-20
ignore phpstorm related configs

devmsh:


jasonmccreary:
This is ignored by my system's global gitignore file. I suggest [creating one](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/).

PR #301, opened 2020-07-20
[WIP] Introduce Token model and Collection

devmsh:
Based on my initial review, I feel that tokens used in many different locations, and It worth to have a Token model and TokenCollection to deal with it.

This PR is just to demonstrate the main idea, and I will complete the refactoring if it fits with your direction.

I have several questions if you initially like the idea:

1. Why we analyze everything and did not pass only and skip to the analyzer.
1. I think token must be able to get themselves as eloquent models do.
1. I don't know if we must pass the blueprint and filesystem from the command down to the Builder then Token model? or we can directly resolve them in the model?
1. What do you prefer between

```php
$tokens = Token::get($blueprint, $files, $draft, [$with_cache=true|false]);

// or

$tokens = Token::get($blueprint, $files, $draft)->withCache();
```

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -2,31 +2,26 @@
 
 namespace Blueprint;
 
+use Blueprint\Models\Token;
 use Illuminate\Filesystem\Filesystem;
 
 class Builder
 {
     public static function execute(Blueprint $blueprint, Filesystem $files, string $draft, string $only = '', string $skip = '')
     {
-        $cache = [];
-        if ($files->exists('.blueprint')) {
-            $cache = $blueprint->parse($files->get('.blueprint'));
-        }
+        $tokens = Token::get($blueprint, $files, $draft);
 
-        $tokens = $blueprint->parse($files->get($draft));
-        $tokens['cache'] = $cache['models'] ?? [];
-        $registry = $blueprint->analyze($tokens);
+        // todo why we analyze everything?
+        $registry = $blueprint->analyze($tokens->toArray());
 
         $only = array_filter(explode(',', $only));
         $skip = array_filter(explode(',', $skip));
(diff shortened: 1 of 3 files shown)


jasonmccreary:
I don't think this is something I will want to refactor at this time.

devmsh:
@jasonmccreary can you direct me to the locations where you would like to refactor the current code? even if it's in the test code?

jasonmccreary:
Sure. Probably the biggest refactor I would want to tackle currently is to centralize a `ModelRegistry`. Nearly all of the _generators_ have some kind of `modelForContext()` method or model array they track internally.

I'd rather this be a separate object passed in to generators, or, more simply, something `Blueprint` tracks and could be asked for (e.g. `Blueprint::modelRegistry()`).

That would remove a lot of duplicate code and make it easier for other generators (or packages) to quickly access the _lexed_ models.

devmsh:
@jasonmccreary based on my initial review, I feel that we need to extract the whole `$tree` and the `Tree` class can deal not only with Models but also with Controllers and other components.

So Generator and other blueprint classes can query the `Tree` 

```php
// nested of 
foreach ($tree['controllers'] as $controller) { // ... }

// we can use
foreach ($tree->controllers() as $controller) { // ... }
// or even
$tree->controllers()->each(function($controller){ // ... })
```

The `Tree` class will be responsible to register its model and provide `modelForContext` itself.

```php
$tree->models()
$tree->modelForContext($context)
```

Waiting for your final thoughts about a separated `ModelRegisty` or a `Tree` class.

jasonmccreary:
Yes. I think a general `Tree` object would be more appropriate to pass to the generators.

PR #303, opened 2020-07-22
extract common logic to abstract generator

devmsh:
By introducing a property for `$new_instance`, we can easily extract the duplicated code between generators and pull them up to an abstract generator class.

Even If we need to customize any of the extracted methods in the future, we can override them so I change the visibility of these methods to protected.

Hope you will find this useful.

I will start working on the `Tree` class as we discussed before, and submit a PR tonight 👍

### DIFF
--- a/src/Generators/StatementGenerator.php
+++ b/src/Generators/StatementGenerator.php
@@ -2,9 +2,9 @@
 
 namespace Blueprint\Generators;
 
-use Blueprint\Contracts\Generator as GeneratorContracts;
+use Blueprint\Contracts\Generator;
 
-abstract class Generator implements GeneratorContracts
+abstract class StatementGenerator implements Generator
 {
     /**
      * @var \Illuminate\Contracts\Filesystem\Filesystem
--- a/src/Generators/Generator.php
+++ b/src/Generators/Generator.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace Blueprint\Generators;
+
+use Blueprint\Contracts\Generator as GeneratorContracts;
+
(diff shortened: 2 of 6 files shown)


jasonmccreary:
I don't understand the purpose of `$new_instance`. Also, if only _statement generators_ extend from this, I'd prefer to make the abstract class more specific, i.e. `StatementGenerator`

devmsh:
@jasonmccreary the `$new_instance` only use to customize the constructor doc block text, I did not see a big advantage from it, but I tried to refactor the code without any breaking changes.

For the 2nd point, the `Generator` currently holds the common `constructor` related logic, but I think it may hold different peace of logic in the next refactoring and some of them may be helpful for other generators?

jasonmccreary:
That's fine to leave `$new_instance`, I'll remove it afterwards.

I understand. However, I am going to change this to `StatementGenerator`. It in turn can extend an abstract `Generator` class in a future refactor if necessary.

devmsh:
Done 👍

jasonmccreary:
Cool. Thanks. Will review and merge tomorrow.

src/Generators/StatementGenerator.php
```diff
@@ -0,0 +1,69 @@
...
+    protected $new_instance = 'new instance';
+
+    public function __construct($files)
+    {
+        $this->files = $files;
+    }
+
+    protected function buildConstructor($statement)
+    {
+        static $constructor = null;
```
jasonmccreary, line 26: @devmsh what was the point of the `static` here? It's causing issues with tests in PHP 8.1.

PR #304, opened 2020-07-22
Introduce Tree class to manage the analyzed tokens

devmsh:
As an initial refactor, I create a new Tree class, and encapsulated the tree management on it.

The class currently provide a central place to provide `modelForContext` and `fqcnForContext` in addition to the simpler getters for `models` `controllers` and `seeders`.

I don't feel comfortable inject the `Tree` object in the method and share it with other methods, and I would suggest having singleton binding or any other pattern to access the Tree from any place in the code.

```php
// nested of 
public function output(Tree $tree): array
{
        $this->tree = $tree;
}
// then
$this->tree->modelForContext($model_name);

we can use
public function output(): array
{
        Tree::controller()
}

Tree::modelForContext($model_name);
```

The PR is already big, so I will wait for your feedback before we can move to the next step.

### DIFF
--- a/src/Blueprint.php
+++ b/src/Blueprint.php
@@ -67,10 +67,10 @@ class Blueprint
             $registry = array_merge($registry, $lexer->analyze($tokens));
         }
 
-        return $registry;
+        return new Tree($registry);
     }
 
-    public function generate(array $tree, array $only = [], array $skip = []): array
+    public function generate(Tree $tree, array $only = [], array $skip = []): array
     {
         $components = [];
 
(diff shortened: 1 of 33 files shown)


jasonmccreary:
I think this is a fine start as is. I'll review and merge in tomorrow's live stream.

devmsh:
Awesome. 

And please give me the next place to focus on so I start reviewing it and share my ideas with you 🔥

jasonmccreary:
That's all I can think of for now. Maybe I'll get more ideas in the live stream.

You are welcome to review the open issues. Maybe pick up an enhancement. 👍

Issue #164, opened 2020-04-30
Custom Stubs for templates?


marbuser:
Hey there,

I was wondering if it was possible we could get different stubs for the views other than blade. For example; Inertia.js, Livewire, Vue components perhaps?, other frameworks I probably don’t know about. 

If this isn’t really feasible because you don’t want to maintain things like that, is it possible we could get some sort of way to define a custom stub? Perhaps provide a path where the files should be created in a config as well as the path to the stub we want to use? 

Fairly new to this package but I’m a big fan and I feel like this is one of the features that will allow more developers of different frameworks to reap the benefits of this amazing package! (Especially since a lot of projects call for something other than blade SSR)

jasonmccreary:
I like it.

Renamed and labelled the issue.

The primary focus of this issue will be for a developer to "publish" the stubs and modify them. Blueprint should attempt to read these first, then fallback to the package versions.

Laravel 7 does something similar. So ideally, we would publish to the same location. Possibly allowing devs to use the same stuff for both Blueprint code generation and `artisan make` generation.

**Note** this would require overlap in the _placeholder_ names used to ensure compatibility. So this is not a requirement, but worth looking into.

n3rdbuddy:
I'd like to second this request. I believe it will add more flexibility to the package and make it way easier to integrate into existing projects.

jasonmccreary:
Great. You, or future readers, please feel free to open PR with some work. Otherwise, I'll probably start working later next week.

jasonmccreary:
@nathane any interest or availability to try and tackle this one?

nathane:
sure, I'll open a pr.

Algeneralo:
@jasonmccreary I'm also looking for the ability to edit other stubs, like controller once and so on
will there be the ability to doing something like this?

jasonmccreary:
We have an open PR for this. I'm hoping to return to it Thursday.

marbuser:
Hey @jasonmccreary

Just wanted to chime back in on this. Since I opened this I’ve been playing around with Livewire and noticed they use an artisan command for their stubs. 

It might be cool not only to be able to define your own custom stubs, but also to disable those stubs completely and just execute an artisan command instead for frameworks that support that already. 

Just another idea I thought of, not sure if you’ve already considered this. 😅

jasonmccreary:
I'm not sure what Livewire does with its stubs or exactly what you're suggesting. Can you provide a real-world use case?

[...]

PR #310, opened 2020-07-24
Fix StatementGenerator

bingtsingw:
I like this feature #164, but instead of committing 100 files at once, I'll split it into several pull requests, every time with simple change.


jasonmccreary:
How does this allow you to publish the _Job_ stub?

bingtsingw:
Forget about publish now, this pr is just a small refactor.
I was modifying stubs to match laravel, and find this refactor can submit independently

jasonmccreary:
Alrighty.

PR #196, opened 2020-05-09
feat: publish stubs

nathane:
This PR allows users to publish all of Blueprint's stubs to `/stubs/blueprint` directory for customization.

Usage:
`php artisan vendor:publish --tag blueprint-stubs`

It currently does not use Laravel 7's stubs. #164

PR #311, opened 2020-07-24
Stub file match Laravel

bingtsingw:
I have replaced all placeholders to `{{ xxx }}` format.
- Changed `Dummyxxx`, `// xxx` to `{{ xxx }}`
- **Renamed** `//`, `//...` to semantic placeholder name. I'm not an English expert, so I'm not sure if they are proper name, maybe need double check.

Some further change discussion:
Do we match stub file structure to [laravel-stubs](https://github.com/jasonmccreary/laravel-stubs), like moving `model/casts.stub` to `model.casts.stub`.

### DIFF
--- a/src/Generators/ControllerGenerator.php
+++ b/src/Generators/ControllerGenerator.php
@@ -73,10 +73,10 @@ class ControllerGenerator implements Generator
 
     protected function populateStub(string $stub, Controller $controller)
     {
-        $stub = str_replace('DummyNamespace', $controller->fullyQualifiedNamespace(), $stub);
-        $stub = str_replace('DummyClass', $controller->className(), $stub);
-        $stub = str_replace('// methods...', $this->buildMethods($controller), $stub);
-        $stub = str_replace('// imports...', $this->buildImports($controller), $stub);
+        $stub = str_replace('{{ namespace }}', $controller->fullyQualifiedNamespace(), $stub);
+        $stub = str_replace('{{ class }}', $controller->className(), $stub);
+        $stub = str_replace('{{ methods }}', $this->buildMethods($controller), $stub);
+        $stub = str_replace('{{ imports }}', $this->buildImports($controller), $stub);
 
         return $stub;
     }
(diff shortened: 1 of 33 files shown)


jasonmccreary:
Probably. The ultimate goal would be if someone has a _published stubs_ in their application, Blueprint would use those. However, I understand that there may be some additional placeholders Blueprint requires or may not know. So I don't expect it to be 100% for every possible customization.

As long as we know the boundaries, I can document them. And, within reason, we can add some _fallback code_ within Blueprint to get around some simple edge cases.

bingtsingw:
Yes, totally agree.
Also, unlike @nathane commented at https://github.com/laravel-shift/blueprint/pull/196#issuecomment-626813760, I think a separate folder `stubs/blueprint` is enough for out goal. Mix laravel's stub with blueprint stub is complex and may cause problems in the future.

jasonmccreary:
For clarity, I think we should mix them where we can within reason. Even if that means we drop something like `{{ body }}` and sniff out the right line in Blueprint.

bingtsingw:
I find it hard to mix them together, not only because blueprint has much more placeholder(it truly does), but they have different approach.

Blueprint split stub into components, and combine them together to make different variant target, Laravel on the other hand, has complete stub files for every command/parameter.

Think of laravel's [controller stubs](https://github.com/jasonmccreary/laravel-stubs/tree/master/stubs), it has `controller.stub`, `controller.plain.stub`, `controller.model.stub`, `controller.model.api.stub`, `controller.nested.stub`, `controller.nested.api.stub` etc. They are all not suitable for blueprint to handle.

Laravel also has some "private" stubs, like `resource.stub`, I don't know why, but they are not published.

In a word, it's really a maintenance burden to compatible with laravel's own stub, and also lead to difficulty usage.

jasonmccreary:
Well, you started a process here. So finish up what you had planned and we'll go from there. Otherwise, I need to revert your contributions.

PR #314, opened 2020-07-25
Add whitelist section to enable code coverage

devmsh:
Hi again!

I enjoyed refactoring the codebase so I can understand the current features before helping introduce a new one or pick a community requested enhancements.

Before going further with extra enhancements, I would like to take care of the `tests` suite, so I enabled the code coverage by adding the whitelisting section to the `phpunit.xml` config file.

The current code coverage reports below show that classes especially the commands and the top-level classes are not well-tested. I just want to double-check if you are interested to cover them?

Screen Shot 2020-07-25 at 16 26 29

In addition to that, I think the `TraceCommand` encapsulate too many concerns and the private method's logic need to be moved to another class to make the command lighter? if yes, what would you name the new class?

Screen Shot 2020-07-25 at 16 33 58

I asked about the `TraceCommand` specifically because it seems to me as the top candidate to be tested and refactored.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -19,6 +19,11 @@
             <directory suffix="Test.php">./tests/Feature</directory>
         </testsuite>
     </testsuites>
+    <filter>
+        <whitelist processUncoveredFilesFromWhitelist="true">
+            <directory suffix=".php">./src</directory>
+        </whitelist>
+    </filter>
     <php>
         <const name="STUBS_PATH" value="stubs"/>
     </php>


devmsh:
BTW, it's not a big concern, but I have no problem to refactor the entire test suite to use PEST if you interested!

jasonmccreary:
Not interested in converting to Pest at this time.

As far as the commands, I agree they need to have more tests. Feel free to start with the primary ones like `build` and we'll work up to `trace`.

devmsh:
Cool, will start with Build Command 👍

PR #315, opened 2020-07-29
BuildCommand tests

devmsh:
In order to test the BuildCommand, I changed some code to make the `Builder` and `BuildCommand` more testable

1. Inject `Builder` to `BuildCommand` nested of Static method invocation to unit test the command. We can also provide e Builder facade and return the static method invocation if needed.
1. Replace `file_exists` with `$this->files->exists` to allow tests code to mock it.
1. Replace `exit(1)` with `return 1` so we can use `assertExitCode(1)` and prevent the command from killing the test command.
1. Simplifying `defaultDraftFile` is not related to the test code, and I can revert the change if you like the old style.

Waiting for your feedback, and if everything goes ok, please select the next candidate to be tested.

Have a nice day 👍

### DIFF
--- a/src/BlueprintServiceProvider.php
+++ b/src/BlueprintServiceProvider.php
@@ -43,7 +43,7 @@ class BlueprintServiceProvider extends ServiceProvider implements DeferrableProv
 
         $this->app->bind('command.blueprint.build',
             function ($app) {
-                return new BuildCommand($app['files']);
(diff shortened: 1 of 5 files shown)


jasonmccreary:
Seems good. I'm going to merge this. But watch for the code style failure next time.

Feel free to continue adding tests for the commands.

PR #317, opened 2020-07-29
Test erase command

devmsh:
Changes:

1. Removing `;` from the `@covers` so code coverage not fails
1. Test the `EraseCommand`

Todo:

- [x] Better way to test if `EraseCommand` calls `TraceCommand`

I tried to use `mock` or `spy` to detect if `TraceCommand` handle method invoked but Laravel neglect the mocked version. Currently I take a very simpler approach to test if `'No models found'` output to the console but it's not a good way to do that.

I also review if Laravel has a similar approach to `Bus::fake($jobsToFake = [])` for commands with no success. If you have any idea, please let me know to I can try to implement it.

### DIFF
--- a/src/Commands/BuildCommand.php
+++ b/src/Commands/BuildCommand.php
@@ -5,7 +5,6 @@ namespace Blueprint\Commands;
 use Blueprint\Blueprint;
 use Blueprint\Builder;
 use Illuminate\Console\Command;
-use Illuminate\Contracts\Filesystem\FileNotFoundException;
 use Illuminate\Filesystem\Filesystem;
 use Illuminate\Support\Str;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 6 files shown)


devmsh:
It already implement that in Laravel Zero [assertCommandCalled](https://github.com/laravel-zero/framework/blob/17723324e1f398b4f7791b8c3a24ebbf4761d0ce/src/Testing/TestCase.php#L53) but I don't know if it worth to use the same approach here.

The 2nd option I think about is to fire some events before and after each command, It may not only help testing if the command is invoked, but also provide some hooks where plugins develop can execute custom logic.

jasonmccreary:
Yeah, this one is tricky. I remember originally trying to test it myself. I'll revisit this in the next live-stream.

Feel free to leave it and more on to another item.

devmsh:
I give it a quick try to replicate an inspired implementation from the `Bus:fake` and it works.

I leave the new changes in a single commit so you can revet it in the live streaming if you want to go with other direction.

if you like the idea, I think I can refactor the PR and move the ArtisanFake to [jasonmccreary/laravel-test-assertions](https://github.com/jasonmccreary/laravel-test-assertions)? and add the package as a composer dependency?

Cann't wait to see you decide in this next live.

PR #323, opened 2020-08-14
Add support for tuple route generation

### LINKED ISSUE #321: Option to generate routes using class/method tuple
- Laravel Version: 7.21.0
- PHP Version: 7.4.3
- Blueprint Version: 1.16.0
- Platform: Linux (Ubuntu 20)

### Issue:

To improve type safety and against accidentally forgetting updating the routes file after renaming a controller class, I disabled the default namespace in the RouteServiceProvider.

However, this is not being picked up by blueprint. The api routes generated only reference the controller(s) via strings - not via class imports - so actually calling the routes fails and has to be corrected manually.

### draft.yaml:
```yaml
models:
  Person:
    id: string
controllers:
  Person:
    resource: api
```

spaceemotion:
I was unsure which approach to take, but I ended up with reading out the actual value from the UrlGenerator class itself, rather than trying to analyze the RouteServiceProvider. Seemed like the most stable solution to me.

Fixes #321

### DIFF
--- a/config/blueprint.php
+++ b/config/blueprint.php
@@ -102,4 +102,17 @@ return [
 
     'use_guarded' => false,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Use Route Tuples
+    |--------------------------------------------------------------------------
+    |
+    | By default, Laravel configures a root controller namespace within the
+    | RouteServiceProvider. In case you have removed that value to use a
+    | "tuple-based" approach in the routes file(s) to improve typing
+    | in your project, you may set this to true instead.
+    |
+    */
+    'use_route_tuples' => false,
+
 ];
--- a/config/blueprint.php
+++ b/config/blueprint.php
@@ -107,10 +107,11 @@ return [
     | Use Route Tuples
     |--------------------------------------------------------------------------
     |
-    | By default, Laravel configures a root controller namespace within the
-    | RouteServiceProvider. In case you have removed that value to use a
-    | "tuple-based" approach in the routes file(s) to improve typing
-    | in your project, you may set this to true instead.
+    | By default, Blueprint follows the Laravel convention of the controller
+    | namespace matches the namespace set within the RouteServiceProvider.
+    | However, you may configure Blueprint to generate routes using a
+    | "tuple syntax" in cases where you may not use this property
+    | or wish to improve static analysis.
     |
     */
     'use_route_tuples' => false,
--- a/config/blueprint.php
+++ b/config/blueprint.php
@@ -114,6 +114,6 @@ return [
     | or wish to improve static analysis.
     |
     */
-    'use_route_tuples' => false,
+    'generate_fqcn_route' => false,
 
 ];
--- a/config/blueprint.php
+++ b/config/blueprint.php
@@ -104,7 +104,7 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Use Route Tuples
+    | Generate FQCN Routes
     |--------------------------------------------------------------------------
     |
     | By default, Blueprint follows the Laravel convention of the controller
--- a/src/Generators/RouteGenerator.php
+++ b/src/Generators/RouteGenerator.php
@@ -6,15 +6,21 @@ use Blueprint\Contracts\Generator;
 use Blueprint\Models\Controller;
 use Blueprint\Tree;
 use Illuminate\Support\Str;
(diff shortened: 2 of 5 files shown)


jasonmccreary:
Per the issue comments, let's make this configurable, not simply based off the `RouteServiceProvider`.

spaceemotion:
@jasonmccreary changed it to use `config()` instead 👍

jasonmccreary:
Much cleaner. 👍 

I'll get this merged once the tests pass and I make some tiny text tweaks. However, I may wait to tag until next week. You can use `dev-master` if you want it right away.

tests/fixtures/routes/routes-tuples.php
```diff
@@ -0,0 +1,5 @@
+
+
+Route::resource('foo', App\Http\Controllers\FooController::class);
```
jasonmccreary, line 3: Unfortunately, this is not (currently) allowed unless the `RouteServiceProvider::$namespace` property is empty.

spaceemotion:
Since this is a configurable value, my guess is that this should be a deliberate action by the user. If they enable it without knowing what it does...

tests/fixtures/routes/routes-tuples.php
spaceemotion: There could be a note about this in the config file.

jasonmccreary:
Hmmm... I considered a PR to the framework to allow the FQCN even with this `RouteServiceProvider` property set. But according to Taylor, it seems as though this property is no longer set by default starting with Laravel 8.

As such, I am comfortable with it as is, and to your point, guiding users to unset the property and/or upgrade to the latest version.

However, we may wish to change the option name to `generate_fqcn_routes` or something...

spaceemotion:
Hm, if Laravel 8 will leave the value blank, the config value shouldn't just be `false` by default then, right? Probably should check `Application::VALUE` if it starts with 7 and disable it.

config/blueprint.php
```diff
@@ -102,4 +102,18 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | By default, Blueprint follows the Laravel convention of the controller
+    | namespace matches the namespace set within the RouteServiceProvider.
+    | However, you may configure Blueprint to generate routes using a
+    | "tuple syntax" in cases where you may not use this property
+    | or wish to improve static analysis.
+    |
+    */
+    'use_route_tuples' => false,
```
spaceemotion, line 117:
```suggestion
    'use_fqn_routes' => \Illuminate\Foundation\Application::VERSION[0] !== '7',
```

src/Generators/RouteGenerator.php
```diff
@@ -52,14 +52,19 @@ protected function buildRoutes(Controller $controller)
         $routes = '';
         $methods = array_keys($controller->methods());
 
-        $className = str_replace('App\Http\Controllers\\', '', $controller->fullyQualifiedClassName());
+        $useTuples = config('blueprint.use_route_tuples');
```
spaceemotion, line 55:
```suggestion
        $useTuples = config('blueprint.use_fqn_routes');
```

config/blueprint.php
jasonmccreary: That's a bit fancy. I worry that will trip devs up when they upgrade their app. Consider that fact that just because newly created Laravel 8 apps will not have it set, all existing apps will.

spaceemotion: Hm, true.

config/blueprint.php
```diff
@@ -102,4 +102,18 @@
 
     'use_guarded' => false,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Use Route Tuples
```
spaceemotion, line 107:
```suggestion
    | Generate FQCN Routes
```

tests/Feature/Generator/RouteGeneratorTest.php
```diff
@@ -88,6 +88,22 @@ public function output_generates_routes_for_mixed_resources()
         $this->assertEquals(['updated' => ['routes/api.php', 'routes/web.php']], $this->subject->output($tree));
     }
 
+    /**
+     * @test
+     */
+    public function output_generates_routes_using_tuples()
+    {
+        config(['blueprint.use_route_tuples' => true]);
```
spaceemotion, line 96:
```suggestion
        config(['blueprint.generate_fqcn_route' => true]);
```

PR #332, opened 2020-08-16
Use aliased resource imports in controllers

spaceemotion:
That way resources do not need to be imported twice: via FQCN and inline import.

(Github will rebase automatically once #330 gets merged)


jasonmccreary:
Looks like we have a conflict after merging #330.

In addition, per our discussion, I'd rather not alias these, but instead just use the "suffixed" name as I do think it's more common.

jasonmccreary:
@spaceemotion, you have any time for this or should I take over?

spaceemotion:
@jasonmccreary that'd be appreciated, as I won't have time until next week. thanks!

PR #337, opened 2020-08-19
Test generator

devmsh:
I'm interested to generate more tests, in my way to achieve this, I do some initial refactoring.

* Add `forcePut` for `FileMixin` will remove the duplicated checks and simplify the test mock preparations.
* Remove some extra parameters
* Extract `TestCase` class and divide the `build` process
* Clean some code using collections nested of standard PHP arrays

Once the `forcePut` approved and have a final name, I will refactor the rest of the project to use it.

The failing test already there before my changes. Seems it happened after merging the lastest PRs, I will take care of it and update the PR after your initial thoughts.


src/Generators/Statements/EventGenerator.php
```diff
@@ -36,7 +36,7 @@ public function output(Tree $tree): array
                     }
 
                     if (! $this->files->exists(dirname($path))) {
-                        $this->files->makeDirectory(dirname($path), 0755, true);
```
jasonmccreary, line 39: Why are all these removed?

devmsh: makeDirectory has only 1 parameter if I'm not wrong?

jasonmccreary: [You're wrong](https://www.php.net/manual/en/function.mkdir.php). I am surprised the tests passed for you when you removed it. Unless you removed it from both sides at the same time. 🙃

jasonmccreary:
I am going to close this. Some of these are incorrect changes, the tests are failing, and generally there is too much in it. Please feel free to open smaller PRs with the individual changes.

Also, quick bit of English, I believe you mean "instead" _instead_ of "nested". 😉 I noticed it here and in your recent issue, so wanted to clairfy.

devmsh:
Will do 👍

Thanks for the note 🥰

PR #339, opened 2020-08-19
Add forcePut in the FileMixin to simplify both code and test expectation

devmsh:


devmsh:
Conflicts resolved

jasonmccreary:
Sorry, but as noted before, I don't see the value in this change. In addition, adding more to `File` creates bloat/confusing on the underlying Laravel application.

Thanks for these PRs, but at this point, I'd rather focus on open issue than refactors.

PR #343, opened 2020-08-25
Small fix for handling id/bigIncrements column with custom names when generating migrations

DennisOng:
For a draft file:
```
  Model:
    custom_column_name: bigIncrements
```

Generated Migrations:
 **Before** 
`$table->id();`

 **After** 
`$table->id('custom_column_name');`

Admittedly not something that's often used

### DIFF
--- a/src/Generators/MigrationGenerator.php
+++ b/src/Generators/MigrationGenerator.php
@@ -127,6 +127,9 @@ class MigrationGenerator implements Generator
             $column_definition = self::INDENT;
             if ($dataType === 'bigIncrements' && $this->isLaravel7orNewer()) {
                 $column_definition .= '$table->id(';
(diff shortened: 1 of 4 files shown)


jasonmccreary:
I'm not keen on this syntax. while `bigIncrements` often denotes the _primary key_, it's not always the case. I'd rather see a new "special" syntax to denote this unconventional change.

For example:

```yaml
model:
    custom_column_name: primary # or pid, pk, etc
```

In addition, denoting a custom primary key name would also have implications for the generated model. So this PR is incomplete.

Given both of these, I am closing. However, if you or someone else address the above, I am willing to add support for this unconventional configuration.

DennisOng:
You're right this affects generated model too, was focused on migrations and missed out on that. No worries and cheers for that detailed explanation!

PR #347, opened 2020-08-26
Generate migration columns with comments

### LINKED ISSUE #344: Need to generate comments for some db columns
### Synopsis:
While generating the migration using laravelshift-blueprint, I want to be able to generate comment for some columns as well.

### Proposed Syntax:
```
models:
  Profession:
    title: string:400
    description: string:400 nullable comment:'Some description for the profession'
```

### Expected Behavior:
```
Schema::create('professions', function (Blueprint $table) {
            $table->id();
            $table->string('title', 400);
            $table->string('description', 400)->nullable()->comment('Some description for the profession');
            $table->timestamps();
        });
```

### Current Output:
```
Schema::create('professions', function (Blueprint $table) {
            $table->id();
            $table->string('title', 400);
            $table->string('description', 400)->nullable();
            $table->timestamps();
        });
```

nathane:
- single-quote and double-quote string accepted.

closes #344

### DIFF
--- a/src/Lexers/ModelLexer.php
+++ b/src/Lexers/ModelLexer.php
@@ -207,7 +207,7 @@ class ModelLexer implements Lexer
         $data_type = null;
         $modifiers = [];
 
-        $tokens = preg_split('#("|\').*?("|\')(*SKIP)(*FAIL)|\s+#', $definition);
+        $tokens = preg_split('#("|\').*?\1(*SKIP)(*FAIL)|\s+#', $definition);
         foreach ($tokens as $token) {
             $parts = explode(':', $token);
             $value = $parts[0];
--- a/src/Lexers/ModelLexer.php
+++ b/src/Lexers/ModelLexer.php
@@ -95,6 +95,7 @@ class ModelLexer implements Lexer
         'primary' => 'primary',
         'foreign' => 'foreign',
         'ondelete' => 'onDelete',
+        'comment' => 'comment',
     ];
 
     public function analyze(array $tokens): array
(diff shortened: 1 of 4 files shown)


tests/fixtures/drafts/columns-with-comments.yaml
```diff
@@ -0,0 +1,4 @@
+models:
+  Profession:
+    title: string:400 comment:"Some title for the profession"
+    description: string:400 nullable comment:'Some description for the profession'
```
spaceemotion, line 4: I thought something like

```suggestion
    title: string:400 "comment:Some title for the profession"
    description: string:400 nullable 'comment:Some description for the profession'
```

should have worked out of the box?

src/Lexers/ModelLexer.php
```diff
@@ -206,7 +207,7 @@ private function buildColumn(string $name, string $definition)
         $data_type = null;
         $modifiers = [];
 
-        $tokens = preg_split('#".*?"(*SKIP)(*FAIL)|\s+#', $definition);
+        $tokens = preg_split('#("|\').*?("|\')(*SKIP)(*FAIL)|\s+#', $definition);
```
jasonmccreary, line 210: We would probably want to ensure the opening and closing quotes match. So maybe the backreference of `\1`?

tests/fixtures/drafts/columns-with-comments.yaml
jasonmccreary: @spaceemotion why is that?

spaceemotion: Based on https://blueprint.laravelshift.com/docs/model-data-types/:

> When specifying an attribute or modifier value which contains a space, you must wrap the value in double quotes ("). For example, enum:Ordered,Completed,"On Hold". Blueprint will unwrap the value during parsing.

I thought this was a general thing for all attributes.

jasonmccreary: I see. It was. Or at least the double quotes. But in your example, you also quoted the _attribute_ (i.e. `comment`)

PR #348, opened 2020-08-26
feat: generate composite indexes for migrations

### LINKED ISSUE #271: Support for composite indexes
Super useful package for quick prototyping. Glad to see this gaining traction and thanks for the effort you've put in.

### Synopsis:
There doesn't seem to be a way to define composite indexes (for migrations) in the draft model files

### Proposed Syntax:
```
models:
    Participation:
        project_id: bigInteger
        user_id: bigInteger
        name: string
        indexes:
          - unique: project_id, user_id
          - index: project_id, user_id
          - primary: project_id, user_id
```
### Expected Behavior:
```
Schema::table('participations', function (Blueprint $table) {
    $table->bigInteger('project_id');
    $table->bigInteger('user_id');
    $table->string('name');
    $table->unique(['project_id', 'user_id']);
    $table->index(['project_id', 'user_id']);
    $table->primary(['project_id', 'user_id']);
});
```

I think the main drawback is introducing the YAML array syntax, since as far as I know no other draft files in Blueprint uses that syntax so far. Otherwise could be quite a handy addition. Happy to take a stab at it if the syntax's approved, thanks!

nathane:
- add composite indexes `_unique`, `_index`, `_primary` (prefixed with `_` to avoid possibly overwriting table names). 
- support multiple definitions of each type with space (` `) between.
- resolves #271 

**draft.yml**
```yaml
models:
  Participation:
    project_id: bigInteger
    team_id: bigInteger
    user_id: bigInteger
    name: string
    _unique: project_id,user_id team_id,project_id
    _index: project_id,user_id
    _primary: project_id,user_id
```
**Output:**
```php
Schema::create('participations', function (Blueprint $table) {
    $table->id();
    $table->bigInteger('project_id');
    $table->bigInteger('team_id');
    $table->bigInteger('user_id');
    $table->string('name');
    $table->unique(['project_id', 'user_id']);
    $table->unique(['team_id', 'project_id']);
    $table->index(['project_id', 'user_id']);
    $table->primary(['project_id', 'user_id']);
    $table->timestamps();
});
```

### DIFF
--- a/src/Blueprint.php
+++ b/src/Blueprint.php
@@ -37,6 +37,8 @@ class Blueprint
             $content = preg_replace('/^(\s*)-\s*/m', '\1', $content);
         }
 
+        $content = preg_replace('/\s*,\s*/', ',', $content);
+
         $content = preg_replace_callback('/^(\s+)(id|timestamps(Tz)?|softDeletes(Tz)?)$/mi', function ($matches) {
             return $matches[1].strtolower($matches[2]).': '.$matches[2];
         }, $content);
--- a/src/Generators/MigrationGenerator.php
+++ b/src/Generators/MigrationGenerator.php
@@ -11,6 +11,12 @@ use Illuminate\Support\Str;
 
 class MigrationGenerator implements Generator
 {
+    const COMPOSITE_INDEXES = [
+        '_index',
+        '_primary',
+        '_unique',
+    ];
+
     const INDENT = '            ';
 
     const NULLABLE_TYPES = [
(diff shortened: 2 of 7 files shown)


jasonmccreary:
Why not use `id` instead of `bigInteger`? Was this just for example, or an actual requirement to using composite keys?

nathane:
I just copied the draft file from the issue, it’s not a requirement.

jasonmccreary:
Cool. Yeah, I'm not a fan of this syntax. I'm going to merge it for the rare cases some might use it. But probably won't document it.

More thinking about improvements for v2. It's possible some kind of `meta` or `_` property is going to be needed for these definitions to define things like _keys_, _relationships_, _policies_, _middleware_.

jasonmccreary:
I decided to move forward with the syntax similar to `relationships` for now, again, in v2 we will likely revisit this syntax by introducing some kind of "meta" key.

mahfoudfx:
Hello,
I'm actually using this syntax and it works perfectly, awaiting for V2. Nevertheless, Ithink that this feature will face another consideration regarding the index name length.
On a project, I had to specify a composite key with 4 columns, I faced the MySQL error 1059 (Identifier name is too long). [https://dev.mysql.com/doc/refman/8.0/en/identifier-length.html] and it cannot be increased on the DBMS.
Currently, a default usage of the command blueprint:build will not handle this case, especially using naming standard for indexes (table.column), the easyiest way is to shorten the table and/or columns names.

jasonmccreary:
@mahfoudfx, that issue is outside the scope of Blueprint.

PR #349, opened 2020-08-27
Add support for generating indexes (including composite indexes) in migrations

DennisOng:
For: https://github.com/laravel-shift/blueprint/issues/271

Unfortunately, the syntax proposed here will require using array within the YAML draft file, which would be incompatible with: https://github.com/laravel-shift/blueprint/pull/260 

Opening this for a discussion, not sure what would be the best way forward to be honest

### DIFF
--- a/src/Blueprint.php
+++ b/src/Blueprint.php
@@ -29,8 +29,8 @@ class Blueprint
         return str_replace('\\', '/', config('blueprint.app_path'));
     }
 
-    public function parse($content, $strip_dashes = true)
-    {
+    public function parse($content, $strip_dashes = false)
+    {   
         $content = str_replace(["\r\n", "\r"], "\n", $content);
 
         if ($strip_dashes) {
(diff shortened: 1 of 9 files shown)


DennisOng:
Ah sorry just saw the other PR from @nathane on this same thing, what a coincidence! 

Feel free to close this off if that works :)

jasonmccreary:
Actually, per the comment I left on that PR, I like your syntax better. Is it possible for you to review that PR and combine any improves (code or test coverage) into this one?

Also, can you update your PR description with the new syntax.

Again, likely something I will change in v2. But the nested array structure is more inline with the current Blueprint syntax than underscores...

jasonmccreary:
Thanks. I have updated this to avoid changing the public method signature of `Blueprint::parse`. Hopefully merging #351 shortly.

DennisOng:
@jasonmccreary Cheers for taking over and making those changes, good idea on limiting the check for array syntaxes, was stuck on the best way forward there.

Sorry for the delayed response to review the initial PR, was having a long overdue `self disconnect --force` weekend :) 

Thanks

PR #356, opened 2020-08-29
Toggle foreign key constraints for Migrations

### LINKED ISSUE #208: [Idea] Separate migration for foreign keys
First of all @jasonmccreary - I'm just blown away by how cool this is. Great work!

---

When generating migrations, it's pretty easy to end up with files that will NOT migrate - because you have foreign keys towards tables that don't exist yet. This is not a Blueprint problem per-se, it's an inconvenience in migrations. But it's a common one, and something that _maybe_ we can fix using Blueprint.

### Synopsis:

Imagine an Articles-Categories-Tags setup for a simple blog:

```yaml
models:
  Article:
    category_id: id foreign # this will fail - there is no Category yet
    title: string
    slug: string unique default:''
    content: text
    image: string nullable
    status: enum:PUBLISHED,DRAFT
    date: date
    featured: boolean default:0
    timestamps: true
    softDeletes: true
    relationships:
      belongsToMany: Tag # this will fail - there is no Tag yet
      
  Category:
    parent_id: id foreign:categories
    left: integer unsigned nullable
    right: integer unsigned nullable
    depth: integer unsigned nullable
    name: string
    slug: string unique
    timestamps: true
    softDeletes: true
    relationships:
      hasMany: Article

  Tag:
    name: string
    slug: string unique
    timestamps: true
    softDeletes: true
    relationships:
      belongsToMany: Article

seeders: Article, Category, Tag

``` 

### Expected Behavior:

Personally, I'd expect the above to work fine - both Models and relationships to be properly defined, and be able to run them successfully.

### Actual Behavior:

With the above, the generated models are perfect. But the migrations can't be run:

Screenshot 2020-05-15 at 10 26 21

Again, the problem is that the **foreign keys point to tables that don't exist yet**. 

There's a quick workaround in the example above: just move `Category` before `Article`, and/or disable foreign keys in Blueprint. This works for this example, but I don't think it'll work for all use cases. If `Category` was dependant on `Tag` too, then there would be no solution for this problem.

### Proposed Syntax:

What if, when generating the migrations, Blueprint just _ignored_ the foreign keys at first pass? Then at the very end, after having generated the migrations for the actual tables, it would generate **a separate migration, only for foreign keys, for all tables**?

The way I see it, this would:
- fix the "_chicken or the egg_" problem;
- require zero syntax change;
- developers won't have to worry about the order their Models are in, which I think makes for a great Developer Experience;

---

What do you think about this @jasonmccreary ? Would this make sense inside Blueprint? Would it be a difficult change?

I think I can dedicate some time in the next 2-3 weeks to understand how Blueprint works and submit a PR, if it's something you don't want to do yourself, but it's something worth doing 😄 

Cheers!

nathane:
- generate migrations with the syntax to disable foreign key constraints when referencing another table.
- supports Laravel 6 & 7 foreign key syntax.
- resolve #208 

```php
class CreateStatesTable extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::disableForeignKeyConstraints();
        Schema::create('states', function (Blueprint $table) {
            $table->id();
            $table->string('name');
            $table->foreignId('countries_id')->constrained('countries')->cascadeOnDelete();
            $table->string('country_code');
            $table->foreign('country_code')->references('code')->on('countries')->onDelete('cascade');
        });
        Schema::enableForeignKeyConstraints();
    }
```

### DIFF
--- a/src/Generators/MigrationGenerator.php
+++ b/src/Generators/MigrationGenerator.php
@@ -37,6 +37,8 @@ class MigrationGenerator implements Generator
     /** @var \Illuminate\Contracts\Filesystem\Filesystem */
     private $files;
 
+    private $hasForeignKeyConstraints = false;
+
     public function __construct($files)
     {
         $this->files = $files;
@@ -90,6 +92,10 @@ class MigrationGenerator implements Generator
         $stub = str_replace('{{ table }}', $model->tableName(), $stub);
         $stub = str_replace('{{ definition }}', $this->buildDefinition($model), $stub);
 
+        if ($this->hasForeignKeyConstraints) {
+            $stub = $this->disableForeignKeyConstraints($stub);
+        }
+
         return $stub;
     }
 
@@ -98,6 +104,10 @@ class MigrationGenerator implements Generator
         $stub = str_replace('{{ class }}', $this->getPivotClassName($segments), $stub);
         $stub = str_replace('{{ table }}', $this->getPivotTableName($segments), $stub);
         $stub = str_replace('{{ definition }}', $this->buildPivotTableDefinition($segments), $stub);
+
+        if ($this->hasForeignKeyConstraints) {
+            $stub = $this->disableForeignKeyConstraints($stub);
+        }
 
         return $stub;
     }
@@ -149,6 +159,7 @@ class MigrationGenerator implements Generator
             $foreign_modifier = $column->isForeignKey();
 
             if ($this->shouldAddForeignKeyConstraint($column)) {
+                $this->hasForeignKeyConstraints = true;
                 $foreign = $this->buildForeignKey(
                     $column->name(),
                     $foreign_modifier === 'foreign' ? null : $foreign_modifier,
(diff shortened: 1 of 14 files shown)


jasonmccreary:
Minor thing, but could you add two line breaks between these lines. It will give the code some breathing room so it's easy to spot these added statements.

Also, could you elaborate on when these are generated?

nathane:
Blueprint will generate the `disable foreign key constraints` syntax when:
1.  a `foreign` modifier is defined on a `Column`.

```YAML
models:
  Comment:
    post_id: foreign
```

2. building a PivotTable with `use_constraints` is set to `true` inside `config/blueprint.php` and the `Column::dataType()` is an `'id'` or `'uuid'`.

```PHP
in_array($column->dataType(), ['id', 'uuid']) && config('blueprint.use_constraints') === true
```

**Output:**
```PHP
// Laravel 6
class CreateCommentsTable extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::disableForeignKeyConstraints();

        Schema::create('comments', function (Blueprint $table) {
            $table->bigIncrements('id');
            $table->unsignedBigInteger('post_id');
            $table->foreign('post_id')->references('id')->on('posts')->onDelete('cascade');
            $table->timestamps();
        });

        Schema::enableForeignKeyConstraints();
    }
```
```PHP
// Laravel 7+
class CreateCommentsTable extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::disableForeignKeyConstraints();

        Schema::create('comments', function (Blueprint $table) {
            $table->id();
            $table->foreignId('post_id')->constrained()->cascadeOnDelete();
            $table->timestamps();
        });

        Schema::enableForeignKeyConstraints();
    }
```

PR #357, opened 2020-08-30
feat: support custom generators via config

### LINKED ISSUE #313: Ability to use custom generators
### Synopsis:
Be able to specify custom generators in `config/blueprint.php` file

### Proposed Syntax:
```
return [
    'generators' => [
         'controller' => \App\CustomBlueprintControllerGenerator::class,
         'model' => \App\CustomBlueprintModelGenerator::class,
         // ...
    ]
];
```

nathane:
- Swap the default generators or Add custom generators using the `config/blueprint.php` file.

resolves #313

### DIFF
--- a/config/blueprint.php
+++ b/config/blueprint.php
@@ -116,4 +116,33 @@ return [
     */
     'generate_fqcn_route' => false,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Generators
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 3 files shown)


jasonmccreary:
I'm open to this, just not sure the syntax. One thing I am curious about is that if this requires everyone to update their config or does it handle things gracefully with the merge?

nathane:
it handles things gracefully now.

PR #362, opened 2020-09-01
Fix psr-4 autoload

nathane:
- rename `Generators` directory
- fix namespace
- exclude `tests/fixtures` from autoloading `.php` stub files.


nathane:
@jasonmccreary I added `exclude-from-classmap` to address this `Deprecation Notice`. https://github.com/laravel-shift/blueprint/pull/363/checks?check_run_id=1067900731#step:8:100

jasonmccreary:
Thanks for the clarification. In the future, please add such comments to you PR (or commit) so it's more clear _why_ you are making these changes.

PR #363, opened 2020-09-01
Update build action

nathane:
- Test Laravel 8
- Update `actions/cache@v2`
- Added `nick-invision/retry@v1` to handle composer timeouts.
- Setup problem matchers for `PHP` & `PHPUnit`

    >This will scan the logs for PHP & PHPUnit errors and warnings, and surface them prominently in the GitHub Actions UI by creating annotations and log file decorations.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,20 +77,42 @@ jobs:
           coverage: none
           ini-values: "memory_limit=-1"
 
+      - name: Setup problem matchers for PHP
+        run: echo "::add-matcher::${{ runner.tool_cache }}/php.json"
+
+      - name: Setup problem matchers for PHPUnit
+        run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
+
       - name: Install highest dependencies from composer.json
         if: matrix.dependencies == 'highest'
-        run: composer config platform --unset && composer update ${{ env.COMPOSER_FLAGS }}
+        uses: nick-invision/retry@v1
+        with:
+          timeout_minutes: 5
+          max_attempts: 5
+          command: composer config platform --unset && composer update ${{ env.COMPOSER_FLAGS }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -58,12 +58,16 @@ jobs:
...
+
       - name: Cache dependencies
-        uses: actions/cache@v1
+        uses: actions/cache@v2
         with:
-          path: ~/.composer/cache/files
+          path: ${{ steps.determine-composer-cache-directory.outputs.directory }}
           key: dependencies-os-${{ matrix.os }}-php-${{ matrix.php-version }}-laravel-${{ matrix.laravel-version }}-composer-${{ hashFiles('**/composer.lock') }}
-          restore-keys: dependencies-os-${{ matrix.os }}-php-${{ matrix.php-version }}-laravel-${{ matrix.laravel-version }}
+          restore-keys: dependencies-os-${{ matrix.os }}-php-${{ matrix.php-version }}-laravel-${{ matrix.laravel-version }}-composer-
```
jasonmccreary, line 70: @nathane, could you explain the change to this line?

nathane: Speed up the cache restoring process by partially matching the `key`. `restore-keys` used for finding the cache if no cache hit occurred for `key`.

> - When there is an exact match, the action restores the files in the cache to the path directory.
> - If there are no exact matches, the action searches for partial matches of the restore keys. When the action finds a partial match, the most recent cache is restored to the path directory.
https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows

jasonmccreary: I see. So technically this was working before, you just added the final `-composer-` to the key.

nathane: yes, that is correct.

PR #365, opened 2020-09-09
Laravel 8 class-based model factories

nathane:
- Add `useLaravel7()`  and `useLaravel8()` environment setup in `TestCase`
- Update `FactoryGenerator`
- Update `ModelGenerator`
- Update `SeederGenerator`
- Backward compatible with Laravel 6 & 7


nathane:
I could be wrong, but I believe the tests are failing because Laravel requires `"php": "^7.3"`

jasonmccreary:
Wow. Thanks for this. I will review more closely and merge in tomorrow's live stream. I'll also work on the seeders as they are now namespaced as well. That shouldn't be as involved of a change though. 😓

PR #367, opened 2020-09-10
Generate tests using Laravel 8 model factories

nathane:
- Generated tests use class-based model factories in Laravel 8


jasonmccreary:
Are we all good here?

nathane:
@jasonmccreary Yes, ready for review.

PR #381, opened 2020-10-02
The destroy test should assert for SoftDeleted

matt-h:
When using a model with `softDeletes`.

The `assertDeleted` check fails with soft deleted models.

This checks if the model has soft deletes and applies the correct assertion.

`draft.yaml` used for testing.
```
models:
  Post:
    title: string
    softDeletes

controllers:
  Post:
    resource
```


jasonmccreary:
Nice one!

PR #387, opened 2020-10-14
fix pivot table naming bug

MuathIbnHassan:
the bug happens when using path prefix on models name

like this one :

```
  Lookup\Region:
    name_en: string:255 nullable
    softDeletes
    relationships:
      hasMany: Lookup\City
```

when there is a belongsToMany relationship, the path will be added to the pivot migration name which is not right


jasonmccreary:
While I appreciate this is the fix, I'd like to see a test for this.

PR #419, opened 2021-01-22
Update phpunit.xml to use current schema

jakub-wolowski:
This fix updates the schema used by PHPUnit. Before applying it, running tests resulted in showing a `Your XML configuration validates against a deprecated schema.` warning.


jasonmccreary:
Technically Blueprint still allows earlier versions of PHPUnit to remain compatible with Laravel 6. As such, I can not merge this into the `master` branch. However, I will use it when we begin work on `v2`.

PR #423, opened 2021-01-28
Add Morphs and uuidMorphs php annotation

### LINKED ISSUE #422: Morphs php annotation incorrect
* Laravel Version: 8.24.0
* PHP Version: 8.0
* Blueprint Version: 1.20.2
* Platform: Mac

In yaml
```yaml
obj: morphs
```

Generated annotation in Model:
```php
/**
 * @property string $obj
 */
```

Expected:
```php
/**
 * @property int $obj_id
 * @property string $obj_type
 */
```

Didn't test morphsUuid...

tanthammar:
fixes #422 
Don't know if you prefer a switch statement or elseif

### DIFF
--- a/src/Generators/ModelGenerator.php
+++ b/src/Generators/ModelGenerator.php
@@ -91,8 +91,20 @@ class ModelGenerator implements Generator
         $phpDoc .= PHP_EOL;
         /** @var Column $column */
         foreach ($model->columns() as $column) {
-            $phpDoc .= sprintf(' * @property %s $%s', $this->phpDataType($column->dataType()), $column->name());
-            $phpDoc .= PHP_EOL;
+            if ($column->dataType() === 'morphs') {
+                $phpDoc .= ' * @property int $'. $column->name() . '_id';
+                $phpDoc .= PHP_EOL;
+                $phpDoc .= ' * @property string $'. $column->name() . '_type';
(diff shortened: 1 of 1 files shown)


jasonmccreary:
`if`/`else` is fine.

Can you back this up with a test? There should be an example fixture already you can add the proper syntax to do demonstrate is generating your new DocBlocks.

tanthammar:
@jasonmccreary 

I don't know the "best practice" for naming the `id` column on the `uuidMorphs` column type.
I've set it to `name._id`, should it be `name.uuid`?

I'll add a test.

jasonmccreary:
@tanthammar I'm not sure what you're asking. Go ahead with the test and I'll review/tweak whatever you send.

nathane:
@tanthammar The [`uuidMorphs()`](https://laravel.com/docs/8.x/migrations#column-method-uuidMorphs) method adds a `{column}_id` `CHAR(36)` equivalent column and a `{column}_type` `VARCHAR` equivalent column.

There is also the [`nullableUuidMorphs()`](https://laravel.com/docs/8.x/migrations#column-method-nullableUuidMorphs) similar to the `uuidMorphs` method; however, the columns that are created will be "nullable". (if you'd like to attempt to add this as well)

PR #415, opened 2021-01-19
Support for migrations ->onUpdate(...)

tanthammar:
This PR add support for migrations `->onUpdate(...)`

Example `draft.yaml`
```yaml
user_id: id nullable foreign onDelete:null onUpdate:null
```

Generates: 
```php
 $table->foreignId('user_id')->nullable()->constrained()->onDelete('set null')->onUpdate('set null');
```

If config is:
```php
    'use_constraints' => true,

    'on_delete' => 'cascade',
    'on_update' => 'cascade',
```

and `draft.yaml` is:
```yaml
user_id: id foreign
```

migration becomes:
```php
$table->foreignId('user_id')->constrained()->cascadeOnDelete()->cascadeOnUpdate();
```

* added option to config  ` 'on_update' => 'cascade',`
* added `'onupdate' => 'onUpdate',` to `ModelLexer`
* updated `MigrationGenerator`

PR #425, opened 2021-01-29
Feature: support onupdate clauses for migrations

nathane:
This PR Supersedes #415

- Added tests to @tanthammar PR.

close #415


tests/fixtures/migrations/foreign-key-on-delete.php
```diff
@@ -17,7 +17,7 @@ public function up()
 
         Schema::create('comments', function (Blueprint $table) {
             $table->id();
-            $table->foreignId('site_id')->constrained()->cascadeOnDelete();
+            $table->foreignId('site_id')->constrained();
```
jasonmccreary, line 20: Why is this one being removed?

nathane: Because that's how it's defined in the draft file. P.S. `config('blueprint.use_constraints') === false `

https://github.com/laravel-shift/blueprint/blob/e966567125b5db484af1acdbd212ebeeb67adcdd/tests/fixtures/drafts/foreign-key-on-delete.yaml#L1-L5

jasonmccreary: Gotcha. I guess I would still want to see a case for `delete` + `cascade` option. I worry we lost some coverage.

nathane: no worries, pushing changes now.

PR #435, opened 2021-02-25
Feature: invokable controllers

jakub-wolowski:
This change allows to properly generate routes for invokable controllers, by defining an `__invoke` method on a controller. 

Considering a definition like the following:
```
controllers:
  Report:
    __invoke:
      render: report
```
      
Before the change, the generated route would be
`Route::get('/report', [App\HttpControllers\ReportController::class, '__invoke']);`
(or `Route::get('/user/{id}', 'ReportController@__invoke']);` without tuples).

After the change, the generated route is
`Route::get('/report', App\HttpControllers\ReportController::class);`
(or `Route::get('/report', 'ReportController');` without tuples).

### DIFF
--- a/src/Blueprint.php
+++ b/src/Blueprint.php
@@ -65,6 +65,10 @@ class Blueprint
             return $matches[1].'resource: web';
         }, $content);
 
+        $content = preg_replace_callback('/^(\s+)invokable?$/mi', function ($matches) {
+            return $matches[1].'invokable: default';
+        }, $content);
+
         $content = preg_replace_callback('/^(\s+)uuid(: true)?$/mi', function ($matches) {
             return $matches[1].'id: uuid primary';
         }, $content);
--- a/src/Blueprint.php
+++ b/src/Blueprint.php
@@ -66,7 +66,7 @@ class Blueprint
         }, $content);
 
         $content = preg_replace_callback('/^(\s+)invokable?$/mi', function ($matches) {
-            return $matches[1].'invokable: default';
+            return $matches[1].'invokable: true';
         }, $content);
 
         $content = preg_replace_callback('/^(\s+)uuid(: true)?$/mi', function ($matches) {
--- a/src/Generators/RouteGenerator.php
+++ b/src/Generators/RouteGenerator.php
@@ -52,11 +52,7 @@ class RouteGenerator implements Generator
         $routes = '';
         $methods = array_keys($controller->methods());
 
-        $useTuples = config('blueprint.generate_fqcn_route');
-
-        $className = $useTuples
-            ? $controller->fullyQualifiedClassName() . '::class'
-            : '\'' . str_replace('App\Http\Controllers\\', '', $controller->fullyQualifiedClassName()) . '\'';
+        $className = $this->getClassName($controller);
 
         $slug = Str::kebab($controller->prefix());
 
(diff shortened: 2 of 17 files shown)


nathane:
@jakub-wolowski Nice work!

```
controllers:
  Post:
    invokable
```

What do you think of this syntax?  Similar to the [`resource`](https://blueprint.laravelshift.com/docs/controller-shorthands/) shorthand.

jakub-wolowski:
Thanks @nathane , in fact a syntax like that was my initial idea, but then I noticed that each statement should be a key: value pair (I must have missed that you can write just `resource` instead of `resource: web`).
So the proposed draft should generate a controller with an empty `__invoke()` method?

jakub-wolowski:
@nathane I've enabled using a `invokable` shorthand that will default to returning a view, so
```
controllers:
  Report:
    invokable
```
is equivalent to
```
controllers:
  Report:
    __invoke:
      render: report
```

You can also still define it explicitly the 'old' way, or use `invokable` instead of `__invoke` in the draft.

I'd be happy to also add this feature's description to the documentation repo if it gets accepted.

nathane:
This is great 👍🏾 but I see room for improvement.

- Why don’t we make the shorthand the default. Meaning, let’s not add the `__invoke` syntax. Using  `invokable` instead of `resource` should automatically generate an invokable controller.

And instead of `invokable: default` use `invokable: true`.

jasonmccreary:
@jakub-wolowski, for clarity, the `invokable` shorthand just returns a view? If so, how is the view name determined? What is the save path?

Also, are both the `__invoke` and `invokable` syntax allowed as well? For example:

```yaml
controllers:
  Report:
    __invoke:
        query:all
        resource: reports
```

nathane:
- I think the `invokable` or `invokable: true` shorthand should return a view by default.
- invokable controllers are "Single Action Controllers"
- Single Action Controllers return an `\Illuminate\Http\Response`, so returning views work.
- view names should be determined, by the English plural version of the controller name. with the `Controller` suffix removed, if found.
- using this `draft.yaml` file below, I think Blueprint should generate these files and updates.
```yaml
// draft.yaml
controllers:
  Report:
    invokable:
        query: all
```
```php
// web.php
Route::get('/report',  \App\Http\Controllers\Report::class);
```
```php
// Report.php - Controller
public function __invoke(): \Illuminate\Http\Response
{
   $reports = \App\Models\Report::all();
    return view('reports', $reports);
}
```
- using the `draft.yaml` file below, I think Blueprint should generate these files and updates.

```yaml
// draft.yaml
controllers:
  Report:
    invokable: true
```
```php
// web.php
Route::get('/report',  \App\Http\Controllers\Report::class);
```
```php
// Report.php - Controller
public function __invoke(): \Illuminate\Http\Response
{
    return view('reports');
}
```
- using the `draft.yaml` file below, I think Blueprint should generate these files and updates. (or maybe this is too much)

```yaml
// draft.yaml
controllers:
  Report:
    invokable:
        save: Report
        // The model name can be omitted if it uses the default name.
```
```php
// web.php
Route::post('/report',  \App\Http\Controllers\Report::class);
```
```php
// Report.php - Controller
public function __invoke(\App\Http\Requests\ReportRequest  $request): \Illuminate\Http\Response
{
    $report = \App\Models\Report::fromRequest($request);
     return view('report', $report);
}
```
- possibly add more options for Single Action Controllers later.
```yaml
// draft.yaml
controllers:
  Report:
    invokable:
        fire: Event
        save: Model
        dispatch: Job
        send: Email
        notify: Notification
        run: Command
```

jasonmccreary:
@nathane, one missing piece, where does the view get saved?

I'm fine with `invokable` being the shorthand and preferred name. However, we should also allow someone to manually define an `__invoke` action and generate an invokable controller and proper route code (to the point of the OP)

nathane:
Views can be stored in the default directory `resources/views` or we can use a custom directory `resources/views/invokable`.

eg. 
`resources/views/reports.blade.php` which we can access with `view('reports');`.
`resources/views/invokable/reports.blade.php` which we can access with `view('invokable.reports');`.

jakub-wolowski:
Right now the `invokable` shorthand is expanded to `invokable: default` in the base `Blueprint` class (I'm OK with changing it to `invokable: true`), which is then changed in `ControllerLexer` to an `__invoke` method that returns a view. The view name is determined by using camel-cased, singular controller's prefix (like `Str::camel(Str::singular($controller->prefix()))`, as it is with resource controller methods). After that, the method definition is processed as usual.

So this is how I see the possible use cases:

- one explicitly defines an `__invoke` method in the draft, and it's treated as any other method definition, but the generated route is correct for an invokable controller
- or one can define the method with the `invokable` keyword to avoid the underscores, the rest behaves the same
- or one can simply use the `invokable` shorthand to generate the default controller that just returns a view, if there's very custom stuff that should happen in the controller

@nathane For the use cases that I encountered a singular version of the controller prefix makes more sense for the view name, but I can make it a plural if you think this would do better.

jasonmccreary:
@jakub-wolowski, if we're going to do it, I'd like to support all 3 bullet points you mentioned. The view template should be generated in the root views folder.

jakub-wolowski:
@jasonmccreary all 3 are supported already in my last commit.
As for the view files, honestly I didn't yet dig into how these are generated, but it should go as with any other method that you define in a draft, as I didn't modify anything related to views generation.

jakub-wolowski:
@jasonmccreary @nathane Do you think anyhing else is needed to make this feature ready to be merged? Just let me know and I'll try to tackle it.

jasonmccreary:
@jakub-wolowski, sorry, I didn't realize you added the final changes from my last comment. I'll try to look at this tomorrow and get it merged.

jakub-wolowski:
@jasonmccreary sure, just let me know if anyhing else could be needed.

PR #437, opened 2021-03-04
Add string limit for faker words

### LINKED ISSUE #434: Model definition using string generates faker words which are too long
- Laravel Version: #.#.#
- PHP Version: 7.4.15
- Blueprint Version: "laravel-shift/blueprint": "^1.22"
- Platform: Windows
- Faker version: "fakerphp/faker": "^1.9.1"

### Issue:
When using pretty short string field definitions, below 8, the seeding fails due to "Database field length exceeded".
I guess this is because the generated factory request creation of faker "word"
`'tag' => $this->faker->word`
in case "string" type was defined in Blueprint.

It would be helpful to have the factory generated in a way which respects the maxLength of the database field.

### draft.yaml:
```yaml
models:
  Post:
    tag: string:3
```

nathane:
```
models:
  Post:
    tag: string:3
```

- `string:{limit}` syntax will now generate `$faker->regexify('[A-Za-z0-9]{limit}')` instead of  `$faker->word`.

Resolves #434

### DIFF
--- a/src/Generators/FactoryGenerator.php
+++ b/src/Generators/FactoryGenerator.php
@@ -245,6 +245,10 @@ class FactoryGenerator implements Generator
                     $faker = 'word';
                 }
 
+                if(($faker === 'word') && (!empty($column->attributes()))){
+                    $faker = sprintf("regexify('[A-Za-z0-9]{%s}')", current($column->attributes()));
+                }
+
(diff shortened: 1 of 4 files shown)


jasonmccreary:
Suprised Faker doesn't have `characters` or `randomLetters`. Would be much more readable than `regexify`.

PR #438, opened 2021-03-04
Add `blueprint:init` console command

### LINKED ISSUE #429: Adding bleuprint:init as alias for blueprint:new
I'd contribute this. It just keeps annoying me as I want to use it like ``git init``.

nathane:
This command is an alias for `blueprint:new`. It creates a `draft.yaml`, if one does not exist, and calls `blueprint:trace`.

closes #429


jasonmccreary:
I feel like this was removed before, but if the masses what it, whatever... 🙃

PR #449, opened 2021-03-23
Refactor filesystem mocks

nathane:
- Mock Filesystem inside `TestCase.php`
- Removed `MocksFilesystem` trait


nathane:
@jasonmccreary this one is ready.

jasonmccreary:
I vaguely remember there being an issue when using `Illuminate\Filesystem\Filesystem` directly. That's why these used the _contract_.

It won't be something that appears in the tests since we're mocking, but it will be an issue at runtime. Especially if someone has swapped it out in the container. However, I'm not sure if we'd want to use _that_ anyway. So it's probably fine, but wanted to brain dump before merging to get your thoughts...

nathane:
- tests were only partially mocking the class. 
- inconsistent usage (concrete/interface)

Blueprint uses both `\Illuminate\Contracts\Filesystem\Filesystem` and `Illuminate\Filesystem\Filesystem` interchangeably.

Which I understand as, it accepts the concrete class in _some_ places instead of the interface, and the tests partially mocks the interface.

jasonmccreary:
@nathane, I'm fine either way. To your point, just want it to be consistent.

src/Generators/Statements/ResourceGenerator.php
```diff
@@ -79,7 +92,7 @@ protected function populateStub(string $stub, Controller $controller, ResourceSt
             . ($controller->namespace() ? '\\' . $controller->namespace() : '');
 
         $stub = str_replace('{{ namespace }}', $namespace, $stub);
-        $stub = str_replace('{{ import }}', $resource->collection() ? 'Illuminate\\Http\\Resources\\Json\\ResourceCollection' : 'Illuminate\\Http\\Resources\\Json\\JsonResource', $stub);
```
jasonmccreary, line 82: These were stringified as they are not available at runtime as part of the Laravel framework.

src/Generators/Statements/ResourceGenerator.php
```diff
@@ -32,9 +43,11 @@ public function output(Tree $tree): array
 
         $output = [];
 
-        $stub = $this->files->stub('resource.stub');
+        $stub = $this->filesystem->stub('resource.stub');
 
-        /** @var \Blueprint\Models\Controller $controller */
+        /**
```
jasonmccreary, line 48: Formatting.

src/Generators/Statements/ResourceGenerator.php
```diff
@@ -8,22 +8,33 @@
 use Blueprint\Models\Model;
 use Blueprint\Models\Statements\ResourceStatement;
 use Blueprint\Tree;
+use Illuminate\Filesystem\Filesystem;
+use Illuminate\Http\Resources\Json\JsonResource;
+use Illuminate\Http\Resources\Json\ResourceCollection;
 use Illuminate\Support\Str;
 
+use function array_diff;
```
jasonmccreary, line 16: No need to import these things. Just noisy.

src/Generators/StatementGenerator.php
```diff
@@ -2,68 +2,88 @@
...
+                ''
+            )
+        );
     }
 
     protected function buildAssignments(array $data)
     {
-        return trim(array_reduce($data, function ($output, $property) {
-            $output .= '        $this->' . $property . ' = $' . $property . ';' . PHP_EOL;
+        return trim(
```
jasonmccreary, line 66: Formatting.

src/Generators/RouteGenerator.php
```diff
@@ -25,17 +39,19 @@ public function output(Tree $tree): array
 
         $routes = ['api' => '', 'web' => ''];
 
-        /** @var \Blueprint\Models\Controller $controller */
+        /**
+         * @var Controller $controller
+         */
         foreach ($tree->controllers() as $controller) {
-            $type = $controller->isApiResource() ? 'api' : 'web';
+            $type           = $controller->isApiResource() ? 'api' : 'web';
```
jasonmccreary, line 46: Formatting.

jasonmccreary (changes requested):
Whatever recent commit you made had very opinionated formatting and importing rules.

jasonmccreary:
@nathane, best to undo the "scouting" commit. It has all sort of opinionated formatting and the class reference, while I generally agree with, actually make the code harder to test since it's not within the Laravel framework.

PR #466, opened 2021-04-09
Fix BC breaks

nathane:
> **How should I handle deprecating functionality?**
> Deprecating existing functionality is a normal part of software development and is often required to make forward progress. When you deprecate part of your public API, you should do two things: (1) update your documentation to let users know about the change, (2) issue a new minor release with the deprecation in place. Before you completely remove the functionality in a new major release there should be at least one minor release that contains the deprecation so that users can smoothly transition to the new API. - [https://semver.org](https://semver.org/#how-should-i-handle-deprecating-functionality)
---
- for projects that require `dbal 2.9`
- ps. `src/Tracer.php` is missing tests 🧨.

### DIFF
--- a/src/Tracer.php
+++ b/src/Tracer.php
@@ -107,7 +107,9 @@ class Tracer
 
             $fetch = method_exists($connection, 'fetchAllAssociative') ? 'fetchAllAssociative' : 'fetchAll';
 
-            $definitions = $connection->$fetch($schema->getDatabasePlatform()->getListTableColumnsSQL($table, $database));
+            $definitions = $connection->{$fetch}(
+                $schema->getDatabasePlatform()->getListTableColumnsSQL($table, $database)
+            );
 
             collect($columns)->filter(function ($column) {
                 return $column->getType() instanceof \Blueprint\EnumType;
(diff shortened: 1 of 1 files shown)


jasonmccreary:
@nathane, not understanding what I'm looking at here. Are you saying this breaks with DBAL 2.9 or that it only works with DBAL 2.9?

nathane:
If a project requires `DBAL 2.9` for their project, and install the latest version of Blueprint, 

Blueprint will try to call `fetchAllAssociative` but that method doesn't exist.

jasonmccreary:
Interesting, I merged that in with #463. I though since `fetchAll` was deprecated that `fetchAllAssociative` already existing in 2.9. I guess not.

jasonmccreary:
`fetchAllAssociative` [is available in 2.x](https://github.com/doctrine/dbal/blob/2.13.x/lib/Doctrine/DBAL/Connection.php#L979). So I'm not sure. Tempted to just revert all this DBAL 3.0 stuff.

jasonmccreary:
@nathane what was the reason for closing?

nathane:
@jasonmccreary It was unnecessary, per your comment regarding `fetchAllAssociative` being available in `DBAL 2.x`.

jasonmccreary:
@nathane sure. I still wonder what these `null` issues are...

nathane:
@jasonmccreary culprit #468

PR #468, opened 2021-04-15
Fix plural-singular modelForContext

### LINKED ISSUE #464: Call to a member function columns() on null
- Laravel Version: 8.12
- PHP Version: 8.0.3
- Blueprint Version: 1.23
- Platform: Mac 

- Issue: 
After running php artisan blueprint:build I get the following error: 
Call to a member function columns() on null
```php

at vendor/laravel-shift/blueprint/src/Generators/TestGenerator.php:251

$setup['data'][] = $faker;
                                 $request_data[$data] = '$' . $variable_name;
                                } else {
                                 foreach ($local_model->columns() as $local_column) {
                                        if ($local_column->name() === 'id') {
                                            continue;
                                        }
```
I found a solution by googling it, but not sure if it's correct or not??? can someone confirm it's the right approach or an error in my yaml file?

```php

 $setup['data'][] = $faker;
                                $request_data[$data] = '$' . $variable_name;
                            } else if (!is_null($local_model)) {
                                foreach ($local_model->columns() as $local_column) {
                                    if ($local_column->name() === 'id') {
                                        continue;
                                    }
 ```
### draft.yaml:
```yaml

```

nathane:
closes #464

eg. the singular of species is species.  so no need to force `WaterSpeciesController` to reference a `WaterSpecy` model. (Currently causing the `null` error)

This will check for and return a `studly` `plural` version of the model name in the `Tree`.

### DIFF
--- a/src/Tree.php
+++ b/src/Tree.php
@@ -41,6 +41,10 @@ class Tree
             return $this->models[Str::studly($context)];
         }
 
+        if (isset($this->models[Str::studly(Str::plural($context))])) {
+            return $this->models[Str::studly(Str::plural($context))];
+        }
+
         $matches = array_filter(array_keys($this->models), function ($key) use ($context) {
             return Str::endsWith($key, '/' . Str::studly($context));
         });


jasonmccreary:
@nathane, nice. More evidence we need some kind of shared _model lookup_ object.

PR #470, opened 2021-04-20
Fix resource pagination

### LINKED ISSUE #439: Error using pagination in index method
- Laravel Version: 8.31.0
- PHP Version: 7.4.3
- Blueprint Version: 1.22.0
- Platform: Linux

### Issue: Error using pagination in index method.
Hello.
When I define a controller specifying the resource type as `resource: api` and in the `index` method it is specified as `query: all` and `resource: 'paginate:posts'`, as shown below:
```yaml
models:
  Post:
    title: string
    body: text
controllers:
  Post:
    resource: api
    index:
      query: all
      resource: 'paginate:posts'
seeders: Post
```

The above code generates the following files:
imagen

The problem comes in the index method of the controller:
```php
public function index(Request $request)
{
    $posts = Post::all();

    return new PostCollection($posts->paginate());
}
```
First it generates a `$posts` variable and assigns it the results of `Post::all()` and then tries to call the `paginate()` method on the `$post` variable. This leads to an error when I make a request to the api/post route:
imagen
```json
{
    "exception": "BadMethodCallException",
    "file": "/home/cristian/Proyectos/blueprint-test/vendor/laravel/framework/src/Illuminate/Macroable/Traits/Macroable.php",
    "line": 103,
    "message": "Method Illuminate\\Database\\Eloquent\\Collection::paginate does not exist.",
    "trace": [
        {
            "class": "Illuminate\\Support\\Collection",
            "file": "/home/cristian/Proyectos/blueprint-test/app/Http/Controllers/PostController.php",
            "function": "__call",
            "line": 22,
            "type": "->"
        }
     ]
}
```

The correct code that should be generated is:
```php
public function index(Request $request)
{
    $posts = Post::paginate();
    
    return new PostCollection($posts);
}
```

### draft.yaml:
```yaml
models:
  Post:
    title: string
    body: text
controllers:
  Post:
    resource: api
    index:
      query: all
      resource: 'paginate:posts'
seeders: Post

```

ghostwriter:
- resolves #439
- `QueryStatement` can now determine the Model from `get` and `pluck` operations.
- No longer a need to define `query: all` when you define `resource: 'paginate:{model}'`.

```diff
# eg. draft.yaml
models:
  Post:
    title: string
    body: text
controllers:
  Post:
    resource: api
    index:
-      query: all
      resource: 'paginate:posts'
seeders: Post
```

### DIFF
--- a/tests/fixtures/drafts/api-resource-pagination.yaml
+++ b/tests/fixtures/drafts/api-resource-pagination.yaml
@@ -1,0 +1,11 @@
+models:
+  Post:
+    title: string
+    body: text
+controllers:
+  Post:
+    resource: api
+    index:
+      query: all
+      resource: 'paginate:posts'
+seeders: Post
--- a/src/Generators/ControllerGenerator.php
+++ b/src/Generators/ControllerGenerator.php
@@ -142,6 +142,20 @@ class ControllerGenerator implements Generator
                     $method = str_replace('* @return \\Illuminate\\Http\\Response', '* @return \\' . $fqcn, $method);
                     $this->addImport($controller, $fqcn);
                     $body .= self::INDENT . $statement->output() . PHP_EOL;
+
+                    if ($statement->paginate()) {
+                        if (! Str::contains($body, '::all();')) {
+                            $queryStatement = new QueryStatement('all', [$statement->reference()]);
+                            $body = implode(PHP_EOL, [
+                                self::INDENT . $queryStatement->output($statement->reference()),
+                                PHP_EOL . $body
+                            ]);
+
+                            $this->addImport($controller, $this->determineModel($controller, $queryStatement->model()));
+                        }
+
+                        $body = str_replace('::all();', '::paginate();', $body);
+                    }
                 } elseif ($statement instanceof RedirectStatement) {
                     $body .= self::INDENT . $statement->output() . PHP_EOL;
                 } elseif ($statement instanceof RespondStatement) {
--- a/src/Models/Statements/QueryStatement.php
+++ b/src/Models/Statements/QueryStatement.php
@@ -26,9 +26,7 @@ class QueryStatement
         $this->operation = $operation;
         $this->clauses = $clauses;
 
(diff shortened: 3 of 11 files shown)


tests/fixtures/drafts/api-resource-pagination.yaml
```diff
@@ -0,0 +1,11 @@
+models:
+  Post:
+    title: string
+    body: text
+controllers:
+  Post:
+    resource: api
+    index:
+      query: all
```
jasonmccreary, line 9: If we don't need to define `query:all` why is it here?

ghostwriter: It was in the original draft provided for the issue, I used it to make sure the issue was first addressed.

jasonmccreary: Is there any value in leaving it? Seems from you PR description, it's no longer a valid combo.

ghostwriter: Yes. Its valid. As in, internally blueprint is “intelligently” defining the 'query: all', so when you ask for a paginated resource it will generate the correct `Model::paginate()`

jasonmccreary: Sorry, I am a little confused. I guess I don't understand this item:

> No longer a need to define `query: all` when you define resource: 'paginate:{model}'.

ghostwriter: I misspoke. Does “shorthand” sound better?

jasonmccreary: Yeah, makes more sense. I guess my ultimate point is, if there is a "shorthand" I just want to ensure there is some kind of test for it. So maybe this fixture could be it?

ghostwriter: Yes. This is an example of the one defining the `query: all`. We already had tests that do not include the query portion.

PR #471, opened 2021-04-21
Adds support for FQCN to model relationships

### LINKED ISSUE #469: [Proposal] Support fully-qualified class names in relationships
First of all: 100% willing to create a complete PR for this, with tests and docs. Just saying 😀

### Synopsis:

When creating Models with relations, we found it's impossible to reference existing models that are in different directories than the configured one (`app` or `app\Models`, whatever):

You can do:
```yaml
models:
  Post:
    title: string:400
    published_at: timestamp nullable
    relationships:
      hasMany: Comment
      belongsToMany: Media, Site
```

But you _can't_ do:
```diff
models:
  Post:
    title: string:400
    published_at: timestamp nullable
    relationships:
-     hasMany: Comment
+     hasMany: Spatie\Comments\Models\Comment
      belongsToMany: Media, Site
```

because the generated relationship will then point to `App\Models\Spatie\Comments\Models\Comment`, which is not what you want... 

The problem happens:
- when you want to point a relationship to a model from a third-party package;
- when you have models both inside `app` and inside `app\Models` (eg. `app\User`, but everything else in `app\Models`);
- when you're doing domain-driven development and your models live in `app\Domains\Store\Models\Comment`;

What the dev has to do in this case is:
- generate the model like this, with the wrong relationship;
- manually edit the file to point to the correct class;
 
Which is... fine, I guess. But I think Blueprint can do better, and we may have an intuitive syntax for it! 😀🤞

### Proposed Syntax:

The _root cause_ we identified is that the developer has no way to tell Blueprint "_I know what I'm doing. Link to this specific class, don't make any assumptions about the namespace_". And the solution we found would be for **Blueprint to also support _fully qualified namespaces_**, as a way to get around the configuration. A simple way to do that would be to... 🥁🥁🥁... **prefix the classname with a backslash**:
- `Comment` means using assumptions, so it'll use `App\Comment` or `App\Models\Comment` depending on the configuration file;
- `\Spatie\Comments\Models\Comment` means I know the _exact_ namespace of my related model, so use that one, no assumptions;

### Expected Behavior:

No change in the examples above. They still generate the same thing.

But this would also work:

```diff
models:
  Post:
    title: string:400
    published_at: timestamp nullable
    relationships:
-     hasMany: Comment
+     hasMany: \Spatie\Comments\Models\Comment
      belongsToMany: Media, Site
```

And the relationship will point to `\Spatie\Comments\Models\Comment::class`, not `App\Models\Spatie\Comments\Models\Comment`.

----

Right now this is the only place we needed the change (relationships), so that's what we've worked to solve. But if you like the `\` syntax... I think we can expand on it in the future, add support for fully-qualified-classnames for Models, Controllers, etc.

@jasonmccreary - **please let me know if you like the `\` syntax. And if you're willing to merge this in the current Blueprint version if we submit a proper PR. Do we have a green light to work on this? Yellow? Red?** 😀 Thanks 🙏

PS. We have a working prototype that we can polish, we'll submit quality PRs, tests, docs - so we'll make it easy for you to merge as a non-breaking change. You don't like something, you ask us to refactor/change, np. We can also promise to maintain that bit of code in the future, for years. But. I totally understand if you don't want Blueprint to support this at all - I'm no stranger to feature creep 🙄 I think this is something worth having, you might not...

tabacitu:
I'm submitting the PR for @pxpm - he's the one who's made all this happen 🙏

Fixes #469 . Allows Blueprint to generate models with relationships to third-party packages (or oddly-placed models).

The developer can type `belongsTo: \App\Models\Cart\Entry` in the YAML file, as the relationship. Notice the `\` prefix - that'll tell Blueprint that it's a fully-qualified class name, so it won't process it any further - it won't prefix `App` or `App\Models` to it, it'll just use it as-is.

### DIFF
--- a/src/Generators/ModelGenerator.php
+++ b/src/Generators/ModelGenerator.php
@@ -190,16 +190,20 @@ class ModelGenerator implements Generator
 
         foreach ($model->relationships() as $type => $references) {
             foreach ($references as $reference) {
+                // if reference starts with \ , we assume developer is using a fully namespaced model
+                $is_model_fqn = Str::startsWith($reference, '\\');
+
                 $custom_template = $template;
                 $key = null;
                 $class = null;
 
                 $column_name = $reference;
-                $method_name = Str::beforeLast($reference, '_id');
+                $method_name = $is_model_fqn ? Str::afterLast($reference, '\\') : Str::beforeLast($reference, '_id');
 
                 if (Str::contains($reference, ':')) {
                     [$foreign_reference, $column_name] = explode(':', $reference);
-                    $method_name = Str::beforeLast($column_name, '_id');
+
+                    $method_name = $is_model_fqn ? Str::afterLast($foreign_reference, '\\') : Str::beforeLast($column_name, '_id');
 
                     if (Str::contains($foreign_reference, '.')) {
                         [$class, $key] = explode('.', $foreign_reference);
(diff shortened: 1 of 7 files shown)


src/Generators/ModelGenerator.php
```diff
@@ -190,51 +190,64 @@ protected function buildRelationships(Model $model)
...
                 }
 
-                $class_name = Str::studly($class ?? $method_name);
-                $fqcn = $this->fullyQualifyModelReference($class_name) ?? $model->fullyQualifiedNamespace() . '\\' . $class_name;
+                // if full model namespace is proviced we will not try to infer it,
+                // we use the namespace as developer gives us
+                if ($is_model_fqn) {
+                    $fqcn = $class ?? $column_name;
+                    $class_name = Str::afterLast($fqcn, '\\');
+                }else{
```
tabacitu, line 226:
```suggestion
                } else {
```

jasonmccreary:
Please fix the style errors and remove all code comments.

src/Generators/ModelGenerator.php
jasonmccreary: @tabacitu I don't think this was submitted correctly...

tabacitu:
Sure, I'll let @pxpm do the fixes since it's his PR and his repo, I don't have write access.

Sidenote: I started refactoring this PR for code reuse - as a start to adding FQCN support for Models, Controllers, etc. That's what I had in mind when I promised a "_cleanly written, fully-tested and well-documented PR_" 😅 but now... I think this PR is _much_ better, since it's less invasive.  

You can find my refactoring in https://github.com/tabacitu/blueprint/pull/2 - but I think I went on cool but wrong path, so I stopped working on it to clear my head. If you _would_ like to add FQCN support in other places in the future and _have time_, give it a look, but... it's not a short read read 😅 (I do that sometimes 😞). If not, don't - I'll probably close it in a few days or weeks when I decide it was a bad idea to begin with.

Cheers!

jasonmccreary:
@tabacitu, I believe the issue is you did not allow the PR to be editable by the maintainer. I'm going to go ahead and merge this. Then I'll make the requested changes.

PR #481, opened 2021-06-11
Pluralize routes

### LINKED ISSUE #480: Allow plural route naming convention
### Synopsis:

First of all, thank you so much for this amazing package! It saves me a ton of time. 

I'm not sure whether this is intentional, but according to Laravel and REST naming conventions, the route URIs should be in plural form.

**Refferences:** 

- https://github.com/alexeymezenin/laravel-best-practices#follow-laravel-naming-conventions
- https://restfulapi.net/resource-naming/
- https://laravel.com/docs/8.x/controllers#resource-controllers

### Expected Behavior:

Currently, the generated routes are in the singular form. For example:

```yml
controllers:
  Post:
    resource: api
```
Generates:
```php
Route::apiResource('post', App\Http\Controllers\PostController::class);
```
Instead, it should be:
```php
Route::apiResource('posts', App\Http\Controllers\PostController::class);
```

pktharindu:
Resolves #480

### DIFF
--- a/src/Generators/RouteGenerator.php
+++ b/src/Generators/RouteGenerator.php
@@ -56,7 +56,7 @@ class RouteGenerator implements Generator
         $routes = '';
         $methods = array_keys($controller->methods());
         $className = $this->getClassName($controller);
-        $slug = Str::kebab($controller->prefix());
+        $slug = Str::kebab(Str::plural($controller->prefix()));
 
         foreach (array_diff($methods, Controller::$resourceMethods) as $method) {
             $routes .= $this->buildRouteLine($className, $slug, $method);


jasonmccreary:
@pktharindu is there any kind of failing test or change to one of the draft file fixtures to prove this change?

pktharindu:
After looking at this carefully, I realize changing this involves a lot more than I initially thought it would. Pluralizing the route URI also pluralizes the route names and the view directories probably need to be pluralized too.

As I'm relatively new to blueprint, attempting to change all that feels like a bit more than I can chew at this point 😄. So, I'm closing this for now.

jasonmccreary:
Indeed. This is something I may make a configuration option in v2 as I know some devs may like to follow a plural convention.

PR #486, opened 2021-07-22
allow non-id columns to be nullable when using foreign

pxpm:
Hello @jasonmccreary hope to find you well.

I think I found an issue with `Blueprint` when using **foreigns**.

So here is the scenario, given the following `yaml` file:

```yaml
models:
  SomeTestsMigration:
    id
    user_id: id foreign:users.id nullable
    something: integer foreign:monsteras.id nullable
    timestamps
```

This would produce the following migration: 

```php
        Schema::disableForeignKeyConstraints();

        Schema::create('some_tests_migrations', function (Blueprint $table) {
            $table->id();
            $table->foreignId('user_id')->nullable()->constrained();
            $table->integer('something'); //notice here its missing the nullable modifier.
            $table->foreign('something')->references('id')->on('monsteras');
            $table->timestamps();
        });

        Schema::enableForeignKeyConstraints();
```

Blueprint does handle the nullable modifier well when the column **dataType** is `id` as you can see here: https://github.com/laravel-shift/blueprint/blob/e446a418c28539f5cd0b6404692dd20930523780/src/Generators/MigrationGenerator.php#L328

After that, it checks and removes nullable from the modifier list as it was already handled: https://github.com/laravel-shift/blueprint/blob/e446a418c28539f5cd0b6404692dd20930523780/src/Generators/MigrationGenerator.php#L210

For `non-id` dataTypes (**something** in my example) it also removes the `nullable` modifier, and it shouldn't because those cases are not handled by Blueprint in prior code.

I've also added tests to prove this use case. 

Let me know if I missed something here.

Best,
Pedro

### DIFF
--- a/src/Generators/MigrationGenerator.php
+++ b/src/Generators/MigrationGenerator.php
@@ -63,10 +63,9 @@ class MigrationGenerator implements Generator
         $tables = ['tableNames' => [], 'pivotTableNames' => []];
 
         $stub = $this->filesystem->stub('migration.stub');
-
         /**
- * @var \Blueprint\Models\Model $model
-*/
+         * @var \Blueprint\Models\Model $model
+         */
         foreach ($tree->models() as $model) {
             $tables['tableNames'][$model->tableName()] = $this->populateStub($stub, $model);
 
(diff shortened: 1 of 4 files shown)


jasonmccreary:
@tabacitu, @pxpm, looks good to me. I'll merge it but wait to tag it for a few days. Feel free to depend upon `dev-master` and let me know everything works.

Issue #490, opened 2021-08-04
[Bug] Wrong namespace on Model relationship


promatik:
- Laravel Version: **8.52**
- PHP Version: **7.4.21**
- Blueprint Version: **dev-master**
- Platform: **Windows**

### Issue:

Hi @jasonmccreary!

I'm using blueprint **dev-master** on my project, and I noticed https://github.com/laravel-shift/blueprint/pull/477 (https://github.com/laravel-shift/blueprint/commit/e446a418c28539f5cd0b6404692dd20930523780 commit) introduced a bug.

The bug happens when you **use blueprint config `models_namespace`**, and you have a `.blueprint` generated with `trace` command.

So to reproduce the issue you'll need to;
1) Set `'models_namespace' => 'Models',` on blueprint config.
2) Run the Tracer command to generate the `.blueprint`
3) The following `draft-yaml`;

### draft.yaml:
```yaml
models:
  Post:
    id
    icon_id: foreign:icons.id
    timestamps
```

This `draft.yaml` will generate the following relationship;

```php
public function icon()
{
    return $this->belongsTo(\App\Models\Models\Icon::class, 'icon_id');
}
```

Note the double `Models` on namespace.

---

Although the problem was introduced on that commit, I believe that commit is not the problem. That commit actually fixes a problem with `modelForContext` function.

What I believe to be the problem is the `Trace` command, because it will wrongly (I assume) generate incorrect paths when `models_namespace` config is set.

This is a `.blueprint` file generated with `Trace` command;

```yaml
models:
    Models\Address: { street: 'string nullable', country: string, icon_id: 'integer nullable', monster_id: integer }
    Models\Icon: { name: string, icon: string }
```

when I run the above yaml, the .blueprint is populated with;

```yaml
models:
    Models\Address: { street: 'string nullable', country: string, icon_id: 'integer nullable', monster_id: integer }
    Models\Icon: { name: string, icon: string }
    Post: { icon_id: 'integer unsigned' }
```

The entries generated with `trace` command are relative to `app`, even when `models_namespace` config is set.
The entries generated with `build` are relative to `models_namespace`.

The entries should be relative to `models_namespace`.

Before #477, since that function was not working there was an "hack" on `Tracer` function `relativeClassName` to avoid this issue, by removing that, everything works as expected. I'll submit a PR to aim that.

tabacitu:
TLDR for the above: 

[...]

PR #491, opened 2021-08-04
Make `relativeClassName` return class name relative to models namespace

promatik:
This is a fix to #490.

This makes `relativeClassName` return the classname relative to models namespace.

The tests are ok, as far as I know this is only used to generate relationships of new entries, and that is working too.

This function was a little bit nonsense, because it first checked if config was set, if so, return, otherwise it would return a string where it would try to replace/remove the config, but there the config was never set anyway. I believe this was introduced to overcome the issue that was fixed on #477.

---

Before this PR, with `'models_namespace' => 'Models'` config, `Trace` would generate the following `.blueprint`;

```yaml
models:
    Models\Address: { street: 'string nullable', country: string, icon_id: 'integer nullable', monster_id: integer }
    Models\Icon: { name: string, icon: string }
```

With this, the paths are relative to `models_namepsace` config;

``` yaml
models:
    Address: { street: 'string nullable', country: string, icon_id: 'integer nullable', monster_id: integer }
    Icon: { name: string, icon: string }
```

This will avoid the "double" `Model\` on relationships namespaces.

### DIFF
--- a/src/Tracer.php
+++ b/src/Tracer.php
@@ -237,9 +237,6 @@ class Tracer
     private function relativeClassName($model)
     {
         $name = Blueprint::relativeNamespace(get_class($model));
-        if (config('blueprint.models_namespace')) {
-            return $name;
-        }
 
         return ltrim(str_replace(config('blueprint.models_namespace'), '', $name), '\\');
     }
--- a/tests/Feature/Commands/TraceCommandTest.php
+++ b/tests/Feature/Commands/TraceCommandTest.php
@@ -47,4 +47,26 @@ class TraceCommandTest extends TestCase
             ->assertExitCode(0)
             ->expectsOutput('Traced 2 models');
     }
+
+    /** @test */
+    public function relative_class_name_removes_models_namespace()
+    {
+        $this->requireFixture('models/comment.php');
+        $this->requireFixture('models/custom-models-namespace.php');
+
+        $method = new \ReflectionMethod(Tracer::class, 'relativeClassName');
+        $method->setAccessible(true);
+
+        // App namespace
+        config(['blueprint.models_namespace' => '']);
+
+        $this->assertEquals($method->invoke(new Tracer(), app('App\Comment')), 'Comment');
+        $this->assertEquals($method->invoke(new Tracer(), app('App\Models\Tag')), 'Models\Tag');
+
+        // Models namespace
+        config(['blueprint.models_namespace' => 'Models']);
+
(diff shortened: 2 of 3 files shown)


tabacitu:
**IMPORTANT! This works, but @jasonmccreary  might consider it a breaking change!** 

---

All tests below are with `'models_namespace' => 'Models'` inside the config. That's the only case where things are different.

On both `v1.24.1` and `master` I've run `php artisan blueprint:trace` and I got:
```php
models:
    Models\Address: { street: 'string nullable', country: string, icon_id: 'integer nullable', monster_id: integer }
    Models\Album: { artist_id: 'biginteger unsigned', name: string, cover: string }
    Models\Artist: { name: string, image: 'string nullable' }
```

After this PR, I've run `php artisan blueprint:trace` and I got:
```php
models:
    Address: { street: 'string nullable', country: string, icon_id: 'integer nullable', monster_id: integer }
    Album: { artist_id: 'biginteger unsigned', name: string, cover: string }
    Artist: { name: string, image: 'string nullable' }
```

Which I know, is the whole point of the PR. But still... **the output of the `trace` command is now different**! 
It _needs_ to be different on `master`, because commit https://github.com/laravel-shift/blueprint/commit/e446a418c28539f5cd0b6404692dd20930523780 only solves half the problem. But still.. it's different.

-----

The scenario where this might be considered a breaking change is when:
- a dev runs the `trace` command
- then switches from `v1.24.1` to `dev-master`
- then does NOT run `trace` again, but tries to do things with the file traced using `v1.24.1`

**I think that scenario is unrealistic, so I wouldn't consider this a breaking change.** I'd rather those very few people have a problem, rather than _everybody using `master`_ which I imagine will be more and more people. So... 

---

I would merge this - **I've just tested it and it works as expected** (noting the output difference above). 

Good find @promatik ! 👏👏👏

jasonmccreary:
@promatik, I'd like to see some kind of test for this. Something to help confirm this won't regress again.

@tabacitu, appreciate the double check, but your _edited_ comment chain is very hard to follow. Consider using ~~strikethrough~~ or leaving individual comments to naturally relay the progression.

jasonmccreary:
@promatik, can you add a test case for this?

promatik:
Hi @jasonmccreary!
Sorry for the delay, I added a test for `relativeClassName()`, the function is private, so I had to use reflection to test it.
Let me know if it's ok.

jasonmccreary:
@promatik, I was looking for a way to test this through a draft file. I don't test private functions directly. However, I already merged it, so I'll see how I might be able to backfill one. 👍

PR #494, opened 2021-08-10
FIX - return types in controller when use resource: api

YannikFirre:
Hello,

When I generate controllers with resource:api and with use_return_types set to true.

The return type controller is set to Response instead of the Resource

Before :
image

After : 
image


jasonmccreary:
@YannikFirre, thank. But I'd like to see a test for this before merging.

YannikFirre:
Hello @jasonmccreary it's done :)

jasonmccreary:
Very nice. Thanks!

PR #495, opened 2021-08-13
ADDED - nested resources

YannikFirre:
Hello,

I have added nested resources when use resource:api.

What do you think about it? 

Good day,

Yannik


jasonmccreary:
I'm not sure what it is. I also would have assumed Laravel automatically used resources for related models.

YannikFirre:
Currently the resource file is only generated with the data of the model and without its relations if these were loaded. 

This addition allows that if they are loaded they are returned when using the resource.

Reference to this: https://laravel.com/docs/8.x/eloquent-resources#relationships

PR #498, opened 2021-08-22
Fix Trace command when non valid files are found

promatik:
Hi @jasonmccreary!

My last PR (https://github.com/laravel-shift/blueprint/pull/488) introduced a possible problem on Trace command, so I'm here to fix it!

We found that non valid files, like a `.zip` file, breaks the Trace command, it generates an empty entry with a backslash (`"\"`), and `class_exists` breaks with that string.

It's a rare situation, but when it happens developers have no feedback about it.
This PR fixes it.

### DIFF
--- a/src/Tracer.php
+++ b/src/Tracer.php
@@ -60,13 +60,17 @@ class Tracer
             $classes = array_merge($classes, $this->filesystem->allFiles($path));
         }
 
-        return array_map(function (\SplFIleInfo $file) {
+        return array_filter(array_map(function (\SplFIleInfo $file) {
             $content = $this->filesystem->get($file->getPathName());
             preg_match("/namespace ([\w\\\\]+)/", $content, $namespace);
             preg_match("/class (\w+)/", $content, $class);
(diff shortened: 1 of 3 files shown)


tabacitu:
@promatik any chance we can have a test for this? It would prove the problem AND make sure it doesn't break in the future... 👀

promatik:
@tabacitu did it 👌

@jasonmccreary, I just added a full test to the Trace command.
It was not possible to test the changed function individually, because it's a private function, it makes only sense to test the whole `Trace` command.

Please note that `Trace` uses the database to retrieve each column information, now to run the tests one must have `sqlite` on their system and the extension must be installed and enable on php – it probably already is, any way here is the command;

```bash
sudo apt-get install php7.4-sqlite
```

jasonmccreary:
Why wouldn't we limit to just `.php` extensions?

promatik:
Did it, makes total sense 👌
And its a performance improvement because the file is not even read.

PR #499, opened 2021-08-23
Update README to reflect new support policy

nexxai:
There was one missed spot where it was said that Laravel 6 is still supported. Wanted to fix that.


jasonmccreary:
Thanks!

PR #500, opened 2021-08-31
change property tree to protected - issue 479 - now you can extend Mo…

rafapaulino:
Change propriety of variable three to protected

In Blueprint\Generators\ModelGenerator; There is a private propriety called $tree - line 18;
When I extends ModelGenerator, I get a error in line 435 because $tree is private.
$model = $this->tree->modelForContext($model_name);


jasonmccreary:
Thanks. I'll merge this later in the week. Feel free to use `dev-master` until then.

PR #503, opened 2021-09-22
If $on is a class, we will try to fetch the table from that class

jyrkidn:
This syntax is currently generating a wrong migration

```yaml
models:
  Event:
    working_title: string
    image_id: uuid foreign:\Foo\Bar\Attachment
```

```php
$table->foreign('image_id')->references('id')->on('\Foo\Bar\Attachments');
```

This makes that inserting record in mysql will fail, since `\Foo\Bar\Attachments` is not an existing table.

The changes in this pull request will fix this.

New output will be 
```php
$table->foreignUuid('image_id')->constrained('attachments');
```

### DIFF
--- a/src/Generators/MigrationGenerator.php
+++ b/src/Generators/MigrationGenerator.php
@@ -300,7 +300,7 @@ class MigrationGenerator implements Generator
         } elseif (Str::contains($on, '.')) {
             [$table, $column] = explode('.', $on);
             $table = Str::snake($table);
-        } elseif (Str::contains($on, '\\') && class_exists($on)) {
+        } elseif (Str::contains($on, '\\')) {
             $table = (new $on())->getTable();
             $column = Str::afterLast($column_name, '_');
         } else {
--- a/src/Generators/MigrationGenerator.php
+++ b/src/Generators/MigrationGenerator.php
@@ -300,6 +300,9 @@ class MigrationGenerator implements Generator
         } elseif (Str::contains($on, '.')) {
             [$table, $column] = explode('.', $on);
             $table = Str::snake($table);
+        } elseif (Str::contains($on, '\\') && class_exists($on)) {
+            $table = (new $on())->getTable();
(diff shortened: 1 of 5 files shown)


jasonmccreary:
@jyrkidn, can you update the PR details with the "fixed" output...

jyrkidn:
Fixed output has been added and tests have been updated after the changes on master.

src/Generators/MigrationGenerator.php
```diff
@@ -294,6 +294,9 @@ protected function buildForeignKey(string $column_name, ?string $on, string $typ
         } elseif (Str::contains($on, '.')) {
             [$table, $column] = explode('.', $on);
             $table = Str::snake($table);
+        } elseif (Str::contains($on, '\\')) {
+            $table = (new $on())->getTable();
```
jasonmccreary, line 298: Depending on the order of operations (i.e. migrations generated before models), it is possible this model will not exist and as such this dynamic creation will fail.

jasonmccreary (changes requested):
I'm wondering if this approach works in all cases...

src/Generators/MigrationGenerator.php
jyrkidn: Thought about that as well, first had a `class_exists` check. Maybe if the class does not exists, we can fallback to a plural, lowercase version of the model? E.g. `\Foo\Bar\Attachment` becomes `attachments`,

jasonmccreary: To limit the code paths, I would not use the dynamic call and just do the plural, lowercase name of the model. I believe this is done elsewhere in the code.

jyrkidn: Updated!

jasonmccreary:
Awesome. Thanks!

jyrkidn:
Thanks for reviewing and merging my pull requests. Very much appreciated!

PR #512, opened 2021-11-17
Append model namespace in the generated factory class.

### LINKED ISSUE #511: Model namespace not appended in the generated model factory class
- Laravel Version: 8.62.0
- PHP Version: 7.4.5
- Blueprint Version: 2.0.0
- Platform: Mac | Windows | Linux

### Issue:
When building a model with a namespace, the generated model factory does not append the provided namespace.

```php
<?php

namespace Database\Factories; // must be: namespace Database\Factories\Admin;
```

### draft.yaml:
```yaml
Admin\User:
    name: string
```

raymadrona:
Resolve #511.

### DIFF
--- a/src/Generators/FactoryGenerator.php
+++ b/src/Generators/FactoryGenerator.php
@@ -83,6 +83,7 @@ class FactoryGenerator implements Generator
             $stub = str_replace('definition()', 'definition(): array', $stub);
         }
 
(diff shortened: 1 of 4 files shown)


jasonmccreary:
@raymadrona, thanks for the issue and subsequent PR. 👍 

I assume the path these factories are being saved is correct? They just didn't have the appropriate namespace?

raymadrona:
The factories are saved correctly on their respective folders, only the namespace is not appropriate.

Thank you also for your amazing work, saves a lot of time.

PR #514, opened 2021-12-10
implemented query::wherenull method

daar:
Hi, I have implemented a new method for the query statement. Please accept my PR


jasonmccreary:
@daar, thanks. Blueprint supports the base level query methods. It will not have a 1:1 syntax with all the Eloquent and Query Builder methods.

PR #516, opened 2021-12-19
feat: remove private in favor of protected

markkimsal:
Making modifications to file layouts that aren't supported by config changes is benefited by subclassing the core generators and changing methods like "getPath".  Using private for every non-contracted method hurts extensibility.


jasonmccreary:
@markkimsal, I'm not sure I'd want to open these up in its current state. At least not without more information on what you're trying to do. Feel free to let me know. Otherwise, closing for now until I can review these internals more.

PR #517, opened 2021-12-27
Bug/remove double white space first relationship method

benjam-es:
When working on a different pull request, I kept hitting an issue, only to discover my auto code cleanup was removing double spaces from my fixtures, which would then prevent the fixture matching the output.

Separated into two commits. One for the fixture cleanup, and another for the solution, incase you want the solution to be carried out in a different way.

Solves #520

### DIFF
--- a/src/Generators/ModelGenerator.php
+++ b/src/Generators/ModelGenerator.php
@@ -164,6 +164,7 @@ class ModelGenerator implements Generator
 
     protected function buildRelationships(Model $model)
     {
+        $newLine = null;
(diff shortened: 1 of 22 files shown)


jasonmccreary:
@benjam-es thanks for this. However, can you work towards a cleaner implementation. The temporary variable, inline coalesce, and iterative assignment are a bit heavy.

benjam-es:
@jasonmccreary much simpler solution added once realising that white space must be trimmed at the end of classes

PR #521, opened 2021-12-27
Update codebase to use default App/Model/Foo structure and namespacing resolves #518

### LINKED ISSUE #518: Update codebase to use newer model namespace/folder structure: app/Models/Foo.php
### Synopsis:
It would make it simpler to contribute if laravel defaults are up to date. With limited codebase knowledge I tried to create a pull request whilst using (copy and pasting) existing tests. It wasn't immediately clear that blueprint does work correctly in the wild, to produce models in the expected namespace, but test cases force change the namespace setting to keep older written tests working

### Proposed Syntax:
For example, change various model instances as follows

``` app/User.php```

to 

``` app/Models/User.php```

### Expected Behavior:
New contributors may naturally use the newest namespace/folder structure, so this would reduce friction

### PR DESCRIPTION
**Pull request by benjam-es**
Addresses #518 

I've added a couple of TODO's to address a couple of possible issues which may need reviewing. I 'made the tests work' but think it may need reviewing.

### CODE REVIEW

File: tests/Feature/Commands/TraceCommandTest.php [hunk 1]
```diff
@@ -62,13 +63,15 @@ public function relative_class_name_removes_models_namespace()
...
 
         // Models namespace
         config(['blueprint.models_namespace' => 'Models']);
 
-        $this->assertEquals($method->invoke(new Tracer(), app('App\Comment')), 'Comment');
-        $this->assertEquals($method->invoke(new Tracer(), app('App\Models\Tag')), 'Tag');
+        $this->assertEquals($method->invoke(new Tracer(), app('App\Models\Comment')), 'Comment');
+        $this->assertEquals($method->invoke(new Tracer(), app('App\Something\Tag')), 'Something\Tag');
+
+        // TODO: Possible bug here. Now using App/Models broke things until I added an extra (original) fixture
```
benjam-es, line 74: First to do.. I believe this could be something within Tracer I didn't quite understand that makes a replacement. So I opted for a different custom fixture using 'Something' instead of Models.

Also may have just been tired, but seems I needed an extra fixture for a comment without Models namespace

tests/Feature/Generators/ModelGeneratorTest.php
```diff
@@ -481,8 +481,8 @@ public function output_generates_models_with_custom_namespace_correctly()
         $path = 'app/Models/Tag.php';
         $model = 'models/custom-models-namespace.php';
 
-        $this->app['config']->set('blueprint.models_namespace', 'Models');
-
+        // $this->app['config']->set('blueprint.models_namespace', '');
+        // TODO: Review this?
```
benjam-es, line 485: I commented out the config for setting namespace because the custom namespace should 'just work' given it matches the new default namespace.

jasonmccreary:
@benjam-es, I am comfortable merging this. A few things though:

1. Resolve the conflicts
2. Remove the `TODO` comments + dead commented code
3. Rebuild the `output_generates_models_with_custom_namespace_correctly` test case to actually use a custom models namespace. For example, `Domain`.

benjam-es:
Completed 1 and 2. If i've understood **3** correctly, that is also complete.

jasonmccreary:
@benjam-es, thanks. What's your Twitter handle, I'll give you a shoutout when I tag the release. 👍

benjam-es:
Twitter: [benjam_es](https://twitter.com/benjam_es)

PR #530, opened 2022-01-02
Add new command to publish stubs

### LINKED ISSUE #528: Stubs/config publish 'improvement'
### Synopsis:
It looks like the current method of publishing the 'stubs' files, is to use ```php artisan vendor:publish --tag=...``` which is usable. I think there may be room for improvement. Similarly it might be useful to complete multiple steps in one line (new, publish stubs, publish config)

### Proposed Syntax:
Use a new command for config publishing, which then simply calls the vendor:publish

```php artisan blueprint:stubs```

The ```blueprint:new``` command could have a couple of flags like so:

```blueprint:new --config```
```blueprint:new --stubs```
```blueprint:new --cs``` (or shorthand)

### Expected Behavior:

1. Specific command for publishing (similar laravel)
2. Streamline 3 steps into 1 step for those wanting to start a new draft file, publish stubs and publish config

benjam-es:
…and add some shortcut flags to the blueprint:new command for streamline publishing - Closes #528

### DIFF
--- a/src/BlueprintServiceProvider.php
+++ b/src/BlueprintServiceProvider.php
@@ -2,16 +2,12 @@
 
 namespace Blueprint;
 
-use Blueprint\Blueprint;
-use Blueprint\Builder;
 use Blueprint\Commands\BuildCommand;
 use Blueprint\Commands\EraseCommand;
-use Blueprint\Commands\NewCommand;
 use Blueprint\Commands\InitCommand;
+use Blueprint\Commands\NewCommand;
+use Blueprint\Commands\PublishStubsCommand;
 use Blueprint\Commands\TraceCommand;
-use Blueprint\Contracts\Generator;
-use Blueprint\FileMixins;
-use Blueprint\Tracer;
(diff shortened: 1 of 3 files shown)


benjam-es:
Added as a separate commit, the ability to simply hook into Laravel's stub:publish command, incase you wanted to make the decision on that separately.

jasonmccreary:
Any reason the shorthand command options need to be capital? Would rather save the keystrokes. Also, I think you can do a single dash (`-`) for short options too ([ref](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Console/ModelMakeCommand.php#L204))

benjam-es:
Happy to change if preferred? 

Just followed the convention in Laravel docs, but I agree, the idea is for the minimal amount of keystrokes.

jasonmccreary:
@benjam-es, do you know if you have to type `--c` or can it just be `-c`?

benjam-es:
the shorthand allow -c, -s and -cs

PR #531, opened 2022-01-04
Improvement: show skipped views in the output of build command

ammardev:
Hello,

This update will print the list of the skipped views while executing `blueprint:build` command. This PR implements the TODO comment in `ViewGenerator:41`.

Before the update:
image

After the update:
image

### DIFF
--- a/src/Commands/BuildCommand.php
+++ b/src/Commands/BuildCommand.php
@@ -96,7 +96,7 @@ class BuildCommand extends Command
     {
         if ($action === 'deleted') {
             return 'error';
-        } elseif ($action === 'updated') {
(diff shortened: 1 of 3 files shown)


jasonmccreary:
@ammardev, in the example, why is `hiiiii.blade.php` being output?

ammardev:
@jasonmccreary 

Because the file already exists.
if the blade view exists, The file is skipped and not recreated.

So it will be printed out to make it clear to the user that the file already exists and not created using Blueprint.

jasonmccreary:
@ammardev, sure. I'm wondering how Blueprint would have generated a `hiiii.php` file to begin with... But I guess it's not important.

Issue #527, opened 2021-12-29
Extract common functionality into an abstract class generator


benjam-es:
### Synopsis:
Extract common functionality into an abstract/base class generator to clean up and simplify code within all generators

### Proposed Syntax:
Some functions will need editing to 'standardise' them before extracting them out such as getPath, buildImports etc. Use blueprint model contract so that we can type hint that in place of the passed blueprint models within common functions.

### Expected Behavior:
Cleaner code, easier to adapt, clean up repeated code

Providing further future benefits
 - Give the model generator access to the buildImports method to reduce the usage of fully qualified class names throughout
 - Cleaner base for a potential switch to the use of blade
 - Improve handling imports in general
   - Allowing new 'default' / configurable default imports per class
   - Allowing for other functionality such as configurable import sorting strategy

Im working on a pull request for this at the moment, so might be worth me keeping a live branch of it for you to look over rather than a large pull request?

jasonmccreary:
@benjam-es, I'm open to cleaning this up. There are several methods which are "copied" between the generators.

However, I am not looking to create multiple class inheritance. Said another way, any PR would need to actually reduce code. Not add a bunch more classes and move the code.

benjam-es:
@jasonmccreary I won't put in a pull request on this one unless requested, as its quite a lot of changes, and some of it may be too opinionated, and/or not how you want things.

Here is a look at reducing around 300+ lines [https://github.com/laravel-shift/blueprint/compare/master...benjam-es:enhancement/refactoring](https://github.com/laravel-shift/blueprint/compare/master...benjam-es:enhancement/refactoring)

It might offer up some suggestions for you. Let me know if you would like me to make it a PR, or sub sections of it

The main bulk of what is done it to separate out some common functionality into an abstract class. Ive also moved some trait/import functionality into separate concerns so it can be reused as needed, with a few other tweaks.

Added benefit of this, is also that models now have access to the addTrait and addImport functionality... this could easily be used for cleaning up FQCN throughout classes

I did start to get into refactoring more of the getPath methods and making models for seeders/factories to make use of just one getPath method, but it started to become a little convoluted. There is also potential for refactoring the output method where some are very similar, but again, this branch has a lot of changes already.

jasonmccreary:
@benjam-es, I am open to this. I do think it's a good streamline. Feel free to open a PR.

A few points I'll make early on though:

- Please mind the code style (Laravel)
[...]

PR #532, opened 2022-01-07
Enhancement/refactoring

benjam-es:
Some streamlining, and extraction of trait and import code, making it usable for model generators. Further discussion on issue #527

### DIFF
--- a/src/Concerns/HandlesImports.php
+++ b/src/Concerns/HandlesImports.php
@@ -6,6 +6,8 @@ use Blueprint\Contracts\Model;
 
 trait HandlesImports
 {
+    protected $imports = [];
+
     protected function addImport(Model $model, $class)
     {
         $this->imports[$model->name()][] = $class;
--- a/src/Concerns/HandlesImports.php
+++ b/src/Concerns/HandlesImports.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Blueprint\Concerns;
+
+use Blueprint\Contracts\Model;
+
+trait HandlesImports
+{
+    protected function addImport(Model $model, $class)
+    {
+        $this->imports[$model->name()][] = $class;
+    }
+
+    protected function buildImports(Model $model)
+    {
+        return collect($this->imports[$model->name()])
+                        ->map(function ($class) {
+                            return "use {$class};";
+                        })
+                        ->unique()
+                        ->sort()
+                        ->implode(PHP_EOL);
+    }
+}
--- a/src/Generators/AbstractClassGenerator.php
+++ b/src/Generators/AbstractClassGenerator.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace Blueprint\Generators;
+
+use Blueprint\Blueprint;
+use Blueprint\Concerns\HandlesImports;
+use Blueprint\Concerns\HandlesTraits;
+use Blueprint\Contracts\Model;
+use Illuminate\Filesystem\Filesystem;
+
+class AbstractClassGenerator
+{
+    use HandlesImports, HandlesTraits;
+
+    const INDENT = '        ';
+
+    protected $filesystem;
+
+    private $tree;
+
+    protected $imports = [];
+    protected $traits = [];
+    protected $stubs = [];
+    protected $output = [];
+    protected $pathSignature = '%s/%s.php';
+
+    public function __construct(Filesystem $filesystem)
+    {
+        $this->filesystem = $filesystem;
+    }
+
+    public function types(): array
+    {
+        return $this->types;
+    }
+
+    protected function getPath(Model $model)
+    {
+        $path = str_replace('\\', '/', Blueprint::relativeNamespace($model->fullyQualifiedClassName()));
+        return sprintf($this->pathSignature, $this->basePath ?? Blueprint::appPath(), $path);
+    }
+
+    protected function create(string $path, $content)
+    {
+        if (!$this->filesystem->exists(dirname($path))) {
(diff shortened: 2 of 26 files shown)


benjam-es:
Seems to be some conflict with the pull request from #531, not sure what that is yet. Will see if I can resolve.

jasonmccreary:
@benjam-es, I believe you use some PHP 7.4+ only features. Hence 7.3 is failing.

I'll probably set a PHP 8.0 requirement later this month once Laravel 9 is released.

benjam-es:
I see i'm still getting them... ill fire docker up with 7.3 and check it out rather than keep trying to push

src/Concerns/HandlesImports.php
```diff
@@ -0,0 +1,22 @@
...
+trait HandlesImports
+{
+    protected function addImport(Model $model, $class)
+    {
+        $this->imports[$model->name()][] = $class;
+    }
+
+    protected function buildImports(Model $model)
+    {
+        return collect($this->imports[$model->name()])
```
jasonmccreary, line 16: Short closure is breaking 7.3 compatibility. Also `unique` and `sort` could be done before map for a minor optimization.

src/Generators/AbstractClassGenerator.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    use HandlesImports, HandlesTraits;
+
+    const INDENT = '        ';
+
+    protected Filesystem $filesystem;
+
+    private Tree $tree;
+
+    protected array $imports = [];
```
jasonmccreary, line 22: Is it possible for these respective property to be added to the `trait` and not the abstract class?

src/Generators/ControllerGenerator.php
```diff
@@ -17,63 +18,30 @@
...
         $stub = $this->filesystem->stub('controller.class.stub');
 
         /** @var \Blueprint\Models\Controller $controller */
         foreach ($tree->controllers() as $controller) {
-            $this->addImport($controller, 'Illuminate\\Http\\Request');
-
+            $this->addImport($controller, \Illuminate\Http\Request::class);
             if ($controller->fullyQualifiedNamespace() !== 'App\\Http\\Controllers') {
-                $this->addImport($controller, 'App\\Http\\Controllers\\Controller');
+                $this->addImport($controller, \App\Http\Controllers\Controller::class);
```
jasonmccreary, line 37: Referencing this via its static `class` property may yield an error if it doesn't exist. Hence the original string method.

Also, aside from this PR, it should probably respect the configured namespaces.

tests/Feature/Lexers/SeederLexerTest.php
```diff
@@ -57,7 +46,7 @@ public function it_returns_seeders()
         $this->assertIsArray($actual['seeders']);
         $this->assertCount(1, $actual['seeders']);
 
-        $this->assertSame(['Post'], $actual['seeders']);
+        $this->assertEquals(['Post'], $actual['seeders']);
```
jasonmccreary, line 49: Why was there a need for the less strict assertion?

jasonmccreary (changes requested):
I'm glad to merge this, provided two things are fixed.

1. Preserve PHP 7.3 compatibility (no short closures).
2. Lessen the generic use of properties and traits (e.g. move respective properties inside trait, only add trait to class that uses it)

jasonmccreary:
@benjam-es, given this a review. Looks good. If you can make those final tweaks, awesome. If not, I'll work on it Tuesday.

tests/Feature/Lexers/SeederLexerTest.php
benjam-es: Believe this was left over from a different change I made and then removed.

src/Generators/AbstractClassGenerator.php
```diff
@@ -0,0 +1,53 @@
...
+
+use Blueprint\Blueprint;
+use Blueprint\Concerns\HandlesImports;
+use Blueprint\Concerns\HandlesTraits;
+use Blueprint\Contracts\Model;
+use Illuminate\Filesystem\Filesystem;
+
+class AbstractClassGenerator
+{
+    use HandlesImports, HandlesTraits;
```
jasonmccreary, line 13: Is it true that all subclasses handle imports? Same for traits?

src/Generators/AbstractClassGenerator.php
```diff
@@ -0,0 +1,53 @@
...
+
+    const INDENT = '        ';
+
+    protected $filesystem;
+
+    private $tree;
+
+    protected $imports = [];
+    protected $traits = [];
```
jasonmccreary, line 22: Can these property not be added to their respective traits?

jasonmccreary (changes requested):
Tests are passing, just the final questions on the genericism.

src/Generators/AbstractClassGenerator.php
benjam-es: It's not true that all subclasses handle imports and traits. It is currently Controller/Seeder/Factory/Test Generators (x4) that make use of the traits, and potentially the 5th with a view to the Model generator starting to use it. So with 5 classes, its around +20 lines (4 per class) and -4 lines from the abstract class.

I was going with the view that, all class generators should have the 'ability' to handle imports and traits, but I take your point, given that Migration, Route and Statement generators don't use this.

I have amended the pull request, along with the other property suggestions

jasonmccreary:
@benjam-es, thanks! I think this is so much cleaner. 💅

Issue #305, opened 2020-07-23
Ability to customize table name


misbahansori:
The default convention is good. but when we want to use a non-English model name, we need to have an ability to specify the name of our table.

jasonmccreary:
I think there's a growing need to have some kind of `meta` property for components. So I am open to this idea if you'd like to propose a syntax.

enricobez:
This feature could be very useful, for example in italian language if I create a model for Utente (User) blueprint will create a table named Utentes that is not italian (Utenti is correct). So there's a way to avoid adding the 's' after the model name?
Thanks

ajcastro:
how about we prefix it with something for example double underscores `__` or at `@` sign?

```yaml
User 
  @table: utenti 
  name: string
  email: string
```

```yaml
User 
  __table: utenti 
  name: string
  email: string
```

ajcastro:
Or maybe we can adapt the class diagram symbols: `#` for protected props, `+` for public, `-` for private

```yaml
User:
  +incrementing: false
  #table: utenti
  -myPrivateProp: "this is a value"
```

jasonmccreary:
This will be something supported in v2. For now, I'm not going to add any more _meta_ columns to Blueprint v1.

Hopefully I'll start v2 in a month or so. Until then, I welcome any additional ideas for this new syntax.

mateoworks:
Is changing table names currently available?
What is the way to use it?

jasonmccreary:
You may change it after the fact. Blueprint is built around Laravel conventions.

PR #533, opened 2022-01-07
fixing add table name on issue #305

kresnayulianta:


jasonmccreary:
@kresnayulianta looks like you're adding new syntax. I may be open to this, however, I would appreciate if you could add a test fixture as well as update the PR to demonstrate the draft syntax.

PR #534, opened 2022-01-13
Respect order for soft deletes. Closes #485

### LINKED ISSUE #485: The softDeletes should be put after the timestamps
- Laravel Version: 8.38.0
- PHP Version: 7.4.16
- Blueprint Version: v1.24.1
- Platform: Windows 

### Issue:

Currently, when I have softDeletes in the yaml, the softDeletes method is put first before the timestamps. Usually 
softDeletes should be put after the timestamps.

Current behavior:

```php
 Schema::create('users', function (Blueprint $table) {
            $table->id();
            $table->string('name');
            $table->softDeletes(); // this should be put after the timestamps
            $table->timestamps();
        });
```

Expected behavior:

```php
 Schema::create('users', function (Blueprint $table) {
            $table->id();
            $table->string('name');
            $table->timestamps();
            $table->softDeletes(); // this should be put after the timestamps
        });
```

I already tried explicitly putting timestamps and softDeletes order in the yaml but still the same output:
```yaml
models:
  User:
    name: string
    timestamps
    softDeletes

```

benjam-es:
Deal with soft deletes in a different way, ensuring migrations respect the order or fields. Closes #485

### DIFF
--- a/tests/Feature/Lexers/ModelLexerTest.php
+++ b/tests/Feature/Lexers/ModelLexerTest.php
@@ -377,7 +377,7 @@ class ModelLexerTest extends TestCase
         $this->assertTrue($model->usesSoftDeletes());
 
         $columns = $model->columns();
-        $this->assertCount(1, $columns);
+        $this->assertCount(2, $columns);
         $this->assertEquals('id', $columns['id']->name());
         $this->assertEquals('id', $columns['id']->dataType());
         $this->assertEquals([], $columns['id']->modifiers());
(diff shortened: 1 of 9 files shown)


tests/Feature/Lexers/ModelLexerTest.php
```diff
@@ -377,7 +377,7 @@ public function it_enables_soft_deletes()
         $this->assertTrue($model->usesSoftDeletes());
 
         $columns = $model->columns();
-        $this->assertCount(1, $columns);
+        $this->assertCount(2, $columns);
         $this->assertEquals('id', $columns['id']->name());
         $this->assertEquals('id', $columns['id']->dataType());
         $this->assertEquals([], $columns['id']->modifiers());
```
jasonmccreary, line 383: Let's add the second column assertion.

jasonmccreary (changes requested):
One small change just to tighten up the assertions now that soft deletes is tracked as a column.

PR #536, opened 2022-01-19
Feature: Inline configuration 

### LINKED ISSUE #529: New draft.yaml syntax - config
### Synopsis:
As a further improvement to the configurability of the package, we could simplify things for those who want to make single simple changes to the configuration often, without needing to publish the full config file and make an edit, then re-run the build command

### Proposed Syntax:
```yaml
models:
  ModelOne:
    column: datatype modifier

  ModelTwo:
    column: datatype
    another_column: datatype modifier

config:
    namespace: MyAppNamespace
    generatePhpdocs: true

````

### Expected Behavior:
Streamline the process for users who often just need simple tweaks. Also provides a visual reference of changes users want with a specific draft file.

This would read the tree for config, and merge the config overrides from the draft yaml file

### PR DESCRIPTION
**Pull request by ghostwriter**
This patch resolves #529 , via the following changes:

- Add tests 
- Create `ConfigLexer` 
- Register `ConfigLexer` in `BlueprintServiceProvider`

#### Note: config keys are case-sensitive.
Default `config/blueprint.php` values.
```yml
config:
    namespace: App
    models_namespace: Models
    controllers_namespace: Http\Controllers
    app_path: app
    generate_phpdocs: false
    use_constraints: false
    on_delete: cascade
    on_update: cascade
    fake_nullables: true
    use_return_types: false
    use_guarded: false
    generators:
      controller: Blueprint\Generators\ControllerGenerator
      factory: Blueprint\Generators\FactoryGenerator
      migration: Blueprint\Generators\MigrationGenerator
      model: Blueprint\Generators\ModelGenerator
      route: Blueprint\Generators\RouteGenerator
      seeder: Blueprint\Generators\SeederGenerator
      test: Blueprint\Generators\TestGenerator
      event: Blueprint\Generators\Statements\EventGenerator
      form_request: Blueprint\Generators\Statements\FormRequestGenerator
      job: Blueprint\Generators\Statements\JobGenerator
      mail: Blueprint\Generators\Statements\MailGenerator
      notification: Blueprint\Generators\Statements\NotificationGenerator
      resource: Blueprint\Generators\Statements\ResourceGenerator
      view: Blueprint\Generators\Statements\ViewGenerator
```

### DIFF
--- a/src/Lexers/ConfigLexer.php
+++ b/src/Lexers/ConfigLexer.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Blueprint\Lexers;
+
+use Blueprint\Contracts\Lexer;
+use Illuminate\Container\Container;
+
+class ConfigLexer implements Lexer
+{
+    private $app;
+
+    public function __construct(Container $app = null)
+    {
+        $this->app = $app ?? Container::getInstance();
+    }
+
+    public function analyze(array $tokens): array
+    {
+        if (array_key_exists('config', $tokens) && is_array($tokens['config'])) {
+            $this->analyzeValue($tokens['config']);
+        }
+        return [];
+    }
+
+    private function analyzeValue(array $config): void
+    {
+        $this->app['config']->set(
+            'blueprint',
+            array_merge(
+                $this->app['config']->get('blueprint'),
+                $config
+            )
+        );
+    }
+}
--- a/src/Lexers/ConfigLexer.php
+++ b/src/Lexers/ConfigLexer.php
@@ -19,6 +19,7 @@ class ConfigLexer implements Lexer
         if (array_key_exists('config', $tokens) && is_array($tokens['config'])) {
             $this->analyzeValue($tokens['config']);
         }
+
         return [];
     }
 
--- a/tests/Feature/Generators/MigrationGeneratorTest.php
+++ b/tests/Feature/Generators/MigrationGeneratorTest.php
@@ -205,6 +205,7 @@ class MigrationGeneratorTest extends TestCase
      */
     public function output_creates_constraints_for_unconventional_foreign_reference_migration()
     {
+        $this->app->config->set('blueprint.use_constraints', true);
         $this->filesystem->expects('stub')
             ->with('migration.stub')
             ->andReturn($this->stub('migration.stub'));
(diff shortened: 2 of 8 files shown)


tests/Feature/Generators/MigrationGeneratorTest.php
```diff
@@ -221,8 +220,7 @@ public function output_creates_constraints_for_unconventional_foreign_reference_
             ->with($model_migration, $this->fixture('migrations/relationships-constraints.php'));
 
         $tokens = $this->blueprint->parse($this->fixture('drafts/relationships.yaml'));
```
jasonmccreary, line 222: I would have expected to see the `yaml` fixture updated.

src/Lexers/ConfigLexer.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        $this->app = $app ?? Container::getInstance();
+    }
+
+    public function analyze(array $tokens): array
+    {
+        if (array_key_exists('config', $tokens) && is_array($tokens['config'])) {
+            $this->analyzeValue($tokens['config']);
+        }
+        return [];
```
jasonmccreary, line 22: Line break before `return` statements. I need to add PHP CS Fixer...

jasonmccreary (changes requested):
This is cool. However, I'd like to see a few fixtures where the `config` section is used to provide more of an integration style tests.

tests/Feature/Generators/MigrationGeneratorTest.php
```diff
@@ -221,8 +220,7 @@ public function output_creates_constraints_for_unconventional_foreign_reference_
...
-        $tree = $this->blueprint->analyze($tokens);
-
+        $tree = $this->blueprint->analyze($tokens + ['config' => ['use_constraints' => true]]);
```
jasonmccreary, line 223: Again, it'd prefer this be pulled from the fixture, not an inline array union.

jasonmccreary (changes requested):

tests/Feature/Generators/MigrationGeneratorTest.php
```diff
@@ -205,6 +205,7 @@ public function output_proper_pascal_case_model_names()
      */
     public function output_creates_constraints_for_unconventional_foreign_reference_migration()
     {
+        $this->app->config->set('blueprint.use_constraints', true);
```
jasonmccreary, line 208: I don't understand why this is needed. Isn't the point of this PR to be able to say `use_constraints: true` in the YAML directly?

tests/fixtures/migrations/relationships.php
```diff
@@ -13,12 +13,16 @@ class CreateCommentsTable extends Migration
      */
     public function up()
     {
+        Schema::disableForeignKeyConstraints();
```
jasonmccreary, line 16: I'm curious why this caused a change in the "after" fixture. I would have expected it to be toggled already given the previous `app->config->set` in the existing test...

ghostwriter: Because the `draft.yml` file for that test fixture was updated to include `config:`, all test that use that same draft file, now update their configuration before generating code.

Prior to this patch, the configuration was only applied explicitly during test setup and generated the same code: 
`/tests/fixtures/migrations/relationships-constraints.php`

PR #542, opened 2022-02-10
chore: remove old version compare code

benjam-es:
Given the package is now php 8.0.0 and higher. Also, those alternate stubs no longer exist


jasonmccreary:
@benjam-es, shouldn't there be corresponding fixtures removed as well?

benjam-es:
When removing the version compare method which conditionally sets the stub name, I searched for the _second_ stubs which were missing. 

I assume that, the code updates that removed those ( < php 8.0.0 ) stubs, must have also removed any relevant tests and fixtures, or some tests would fail?

jasonmccreary:
Interesting.

PR #544, opened 2022-02-11
Default to plural api route naming, with config to allow plural or singular

### LINKED ISSUE #480: Allow plural route naming convention

benjam-es:
Fixes #480

Issue created requiring plural api resource slugs. Hopefully this covers the issue as you requested @pktharindu?

This implementation seems very simple, so there may be more needed, so happy to take any pointers @jasonmccreary or @pktharindu

### DIFF
--- a/config/blueprint.php
+++ b/config/blueprint.php
@@ -116,6 +116,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Pluralize base slug for Route::apiResource(...)
+    |--------------------------------------------------------------------------
+    |
+    | By default, Blueprint will pluralize the api resource slug meaning
+    | a model named post will create an api resource route definition with a
+    | the base slug as 'posts'. Previously this was singular. Setting this
+    | to false, will product a singular slug of 'post'
+    |
+    */
+    'plural_api_slug' => true,
+
+    /*
+    |--------------------------------------------------------------------------
     | Generators
     |--------------------------------------------------------------------------
     |
--- a/src/Generators/RouteGenerator.php
+++ b/src/Generators/RouteGenerator.php
@@ -53,7 +53,7 @@ class RouteGenerator extends AbstractClassGenerator implements Generator
         $resource_methods = array_intersect($methods, Controller::$resourceMethods);
         if (count($resource_methods)) {
             $routes .= $controller->isApiResource()
-                ? sprintf("Route::apiResource('%s', %s)", $slug, $className)
+                ? sprintf("Route::apiResource('%s', %s)", config('blueprint.plural_api_slug') ? Str::plural($slug) : Str::singular($slug), $className)
                 : sprintf("Route::resource('%s', %s)", $slug, $className);
 
             $missing_methods = $controller->isApiResource()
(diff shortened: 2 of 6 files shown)


jasonmccreary:
It really depends on how deep the developer feels the plural should go: route? controller name? view folder name? etc...

In this end, this is one developers opinion. As such, it's unlikely I would merge this without some consensus.

benjam-es:
My reading of issue #480 is that he is referencing the widely used api resource naming convention using plural naming. So in this instance, I think it stands alone, and the pull request meets that one request only (with added config so its not strictly enforced).

pktharindu:
While I have seen devs use both plural and singular resource routes, I believe the plural form is more common and is in accordance with [Laravel docs](https://laravel.com/docs/controllers#api-resource-routes) and REST naming conventions.

The only other thing I usually pluralize is the view folder name. But I'm not sure what the norm is in that regard. So I agree with Jason that we need some consensus.

jasonmccreary:
I think the option for routes is good. It's really a matter if we want to also pluralize the views folder.

benjam-es:
I think @pktharindu's raised issue (#480) offers a better sensible default for api resources, so think the pull request has merit for that reason. 

I'm happy to amend or delete if needed.

jasonmccreary:
Agreed. I will merge this as-is. However, trying to ask if either of you feel we should pluralize the views folder when this option is enabled?

pktharindu:
View folders are usually named after the routes, which makes them easier to find IMHO. Package like [spatie/laravel-route-discovery](https://github.com/spatie/laravel-route-discovery) seems to follow the same convention. So, I think it is a sensible thing to do so. What do you guys think?

dalpat:
Yes pls merge this, it's the only reason I'm unable to use this package. It's great what this package has to offer but forcing singular is a big entry barrier for me as all my projects follow the plural naming conventions for routes and view folders.

jasonmccreary:
I'm going to be working on Blueprint in [my live streams](https://www.youtube.com/playlist?list=PLmwAMIdrAmK5rH3SWvokHV8xI_0mauxDL) during the month of October. I will start with this.

dalpat:
That would be wonderful.

spekulatius:
> I'm going to be working on Blueprint in [my live streams](https://www.youtube.com/playlist?list=PLmwAMIdrAmK5rH3SWvokHV8xI_0mauxDL) during the month of October. I will start with this.

Nice, subscribed :muscle:

jasonmccreary:
Given the age, I am closing. However, I have used this work to create #574.

Issue #548, opened 2022-02-17
Laravel 9, Scout fullText index


tanthammar:
### Synopsis:
Laravel 9 comes with a database driver for Laravel Scout. It would be great if we could add support for `fullText` indexes which is one of the recommended data types.

### Proposed Syntax:

`foo: fullText` should output `$table->fullText('foo')` in migrations

### Expected Behavior:

I hoped it was enough to add `'fullText' => 'fullText', ` to the `ModelLexer->dataTypes` array, but it didn't do it :)

jasonmccreary:
I like it. Feel free to take a stab at a PR.

PR #550, opened 2022-02-18
Laravel 9, Scout fullText index

tanthammar:
As discussed in #548 I am trying to add support for fullText indexes.

### Synopsis:
Laravel 9 comes with a database driver for Laravel Scout. It would be great if we could add support for `fullText` indexes which is one of the recommended data types.

### Proposed Syntax:

`foo: fullText` should output `$table->fullText('foo')` in migrations

I don't know which files to change but I tried :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .env
 .phpunit.result.cache
 composer.lock
+/.idea
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@
 .env
 .phpunit.result.cache
 composer.lock
-/.idea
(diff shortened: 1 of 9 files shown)


jasonmccreary:
@tanthammar, this looks really good. Two quick things:

1. Please remove the `.gitignore` change as it's unrelated.
2. I'd like to see a test/fixture to prove the migration changes.

tanthammar:
Something like this? :) Added fixtures/migrations/full-text.php

jasonmccreary:
Yes. But can you also add it to the `MigrationGeneratorTest` it runs as part of the test.

tanthammar:
Sorry, give up. Don't know what I'm supposed to do :)

jasonmccreary:
@tanthammar, ok. I'll finish it out later.

PR #559, opened 2022-03-23
Added newFactory() to models #558

arnonm:


jasonmccreary:
I am not going to merge the `strict_types` PR and it seems this one builds off that. In addition, one of Blueprint's core goals is to leverage Laravel conventions. As such, automatically creating the `newFactory` method for all models does not align with that goal.

PR #560, opened 2022-03-23
Added declare(strict types=1) to all generated files as an option. Default flag status is false

arnonm:


jasonmccreary:
@arnonm, thanks for this.

If you could please fix the styling issues and, ideally, write a test where you toggle the new option and assert it added `declare(strict_types=1)`.

Issue #546, opened 2022-02-16
SoftDeletes Issue


shepherdscribe:
- Laravel Version: 8.0
- PHP Version: 8.0
- Blueprint Version: 2.2
- Platform: Mac

### Issue:
softDeletes issue on model factory still treats it as string.
<img width="814" alt="Screen Shot 2022-02-16 at 11 04 33 AM" src="https://user-images.githubusercontent.com/62448059/154188301-4d913ff1-522c-4c9e-905b-1ee844af3965.png">

### draft.yaml:
```yaml
# PLEASE PASTE THE DRAFT FILE WHICH REPRODUCES THE ISSUE HERE...
```
Address:
    client_id: unsignedBigInteger foreign:clients
    street: string
    region_id: unsignedInteger
    province_id: unsignedInteger
    city_id: unsignedInteger
    barangay_id: unsignedInteger
    zip_code: unsignedInteger min:3 max:4
    user_id: unsignedBigInteger nullable foreign:users
    type: enum:primary,billing,user default:primary
    softDeletes

mgkimsal:
FWIW, it's still doing this on blueprint 2.3 (as of today)

jasonmccreary:
Very strange. Seems the factory is keeping a literal "softdeletes" column name. Open to PRs. Otherwise, this will probably need to wait until June.

mgkimsal:
@jasonmccreary https://github.com/mgkimsal/blueprint/commit/1d5ca05a51aa295165b2aa84fee04240d7156331

haven't done a specific PR because... the ModelLexerTest breaks

```php
 $model = $actual['models']['Model'];
        $this->assertEquals('Model', $model->name());
        $this->assertTrue($model->usesTimestamps());
        $this->assertTrue($model->usesSoftDeletes());

        $columns = $model->columns();
        $this->assertCount(2, $columns);
        $this->assertEquals('id', $columns['id']->name());
        $this->assertEquals('id', $columns['id']->dataType());
        $this->assertEquals('softdeletes', $columns['softdeletes']->name());
        $this->assertEquals('softDeletes', $columns['softdeletes']->dataType());
        $this->assertEquals([], $columns['id']->modifiers());
```

[...]

PR #564, opened 2022-05-21
fix: two pass filter to remove softDeletes - issue 546

mgkimsal:
Soft delete columns aren’t ‘fillable’ so shouldn’t be part of the generated factory class.

Second pass at fixing issue #546 

No test changes needed.

### DIFF
--- a/src/Generators/FactoryGenerator.php
+++ b/src/Generators/FactoryGenerator.php
@@ -194,9 +194,14 @@ class FactoryGenerator extends AbstractClassGenerator implements Generator
             return $columns;
         }
 
-        return array_filter(
+        $nonNullableColumns = array_filter(
             $columns,
             fn (Column $column) => !in_array('nullable', $column->modifiers())
+        );
+
+        return array_filter(
+            $nonNullableColumns,
+            fn (Column $column) => $column->dataType() !== 'softDeletes'
         );
     }
 


jasonmccreary:
@mgkimsal, thanks. It does appear these tests are passing. What's interesting to me is that none of the existing tests were failing. I'd like to think one of them would have demonstrated this issue.

I'd like to better understand it before merging this.

mgkimsal:
> @mgkimsal, thanks. It does appear these tests are passing. What's interesting to me is that none of the existing tests were failing. I'd like to think one of them would have demonstrated this issue.

I was somewhat confused by that as well, and... I couldn't verify in the test runner that that code was definitely getting exercised the same way as running it 'live' was.

Again, I'm not an expert in how all this is put together.  I can verify the problem, and this fix works, and ideally you can see the problem too.  If this helps you find the root and you can fix it another way, that great as well.  I don't particularly care how it gets fixed, just trying to help.

Thanks.

jasonmccreary:
I believe your solution works. I guess I'm curious if this is the only issue (just factories), or if something is flawed with how `SoftDeletes` columns are tracked.

I may play with this is a bit before merging as I'd like to have seen a failing test, then it passing with your solution.

Thanks.

jasonmccreary:
Unfortunately when I add a test case for this and run it locally, I see the failure mentioned in #546.

Closing and going to push your changes to another PR as GitHub doesn't make it easy to add to a collaborator's PR.

PR #569, opened 2022-08-19
Update migration.stub

sajaddp:
Hi,
A very small update to match the file structure with the default migrations structure in Laravel


jasonmccreary:
Nice. I thought I did this a while back, but guess not.

Unfortunately it causes all the tests to fail. You're welcome to fix this and reopen.

PR #576, opened 2022-10-06
Remove double "for"

spekulatius:
Hey @jasonmccreary,

just watched your stream from yesterday and noticed you had a double *for* in the comment block :+1: 

Thanks for building Blueprint! :muscle:

Cheers,
Peter

### DIFF
--- a/config/blueprint.php
+++ b/config/blueprint.php
@@ -119,7 +119,7 @@ return [
     | Pluralize route names
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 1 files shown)


jasonmccreary:
Thanks. Unfortunately it doesn't seem like you allowed maintainers to edit. As such, I can't tweak the cascaded wording. So I'll make this change on `master`.

spekulatius:
All good, just wanted to fix it. Classical case of "Code-OCD"

PR #584, opened 2022-12-13
Fixing missing `use` in generated tests for show-only controllers (fi…

### LINKED ISSUE #583: Missing import on single-action controller
- Laravel Version: v9.43.0
- PHP Version: PHP 8.1.13
- Blueprint Version: v2.4.0
- Platform: ArchLinux

### Issue:

It seems the generated controller test misses the `use` for the related model, if only a `show`-action is defined on the controller.

### draft.yaml:
```yaml
models:
  Entry:
    slug: string:400
    title: string:400
    intro: longtext

controllers:
  Entry:
    show:
      render: entry.show with:entry
```

Here the generated ControllerTest with the expected diff:

```diff
<?php

namespace Tests\Feature\Http\Controllers;

+use App\Models\Entry;
use Tests\TestCase;

/**
 * @see \App\Http\Controllers\EntryController
 */
class EntryControllerTest extends TestCase
{
    /**
     * @test
     */
    public function show_displays_view()
    {
        $entry = Entry::factory()->create();

        $response = $this->get(route('entry.show', $entry));

        $response->assertOk();
        $response->assertViewIs('entry.show');
        $response->assertViewHas('entry');
    }
}

```

spekulatius:
Here is how I would fix #583.

### DIFF
--- a/src/Generators/TestGenerator.php
+++ b/src/Generators/TestGenerator.php
@@ -110,6 +110,8 @@ class TestGenerator extends AbstractClassGenerator implements Generator
                 $setup['data'][] = sprintf('$%s = %s::factory()->create();', $variable, $model);
             }
 
+            $this->addImport($controller, $modelNamespace . '\\' . $model);
+
             foreach ($statements as $statement) {
                 if ($statement instanceof SendStatement) {
                     if ($statement->isNotification()) {
@@ -393,7 +395,6 @@ class TestGenerator extends AbstractClassGenerator implements Generator
                     $this->addRefreshDatabaseTrait($controller);
 
                     $model = $this->determineModel($controller->prefix(), $statement->reference());
-                    $this->addImport($controller, $modelNamespace . '\\' . $model);
 
                     if ($statement->operation() === 'save') {
                         $tested_bits |= self::TESTS_SAVE;
@@ -639,8 +640,6 @@ END;
         }
 
         $faker = sprintf('$%s = %s::factory()->create();', Str::beforeLast($local_column->name(), '_id'), Str::studly($reference));
-
-        $this->addImport($controller, $modelNamespace . '\\' . Str::studly($reference));
 
         return [$faker, $variable_name];
     }
--- a/src/Generators/TestGenerator.php
+++ b/src/Generators/TestGenerator.php
@@ -107,10 +107,10 @@ class TestGenerator extends AbstractClassGenerator implements Generator
                 : config('blueprint.namespace');
 
             if (in_array($name, ['edit', 'update', 'show', 'destroy'])) {
+                $this->addImport($controller, $modelNamespace . '\\' . $model);
+
                 $setup['data'][] = sprintf('$%s = %s::factory()->create();', $variable, $model);
             }
-
-            $this->addImport($controller, $modelNamespace . '\\' . $model);
 
             foreach ($statements as $statement) {
                 if ($statement instanceof SendStatement) {
@@ -395,6 +395,7 @@ class TestGenerator extends AbstractClassGenerator implements Generator
                     $this->addRefreshDatabaseTrait($controller);
 
                     $model = $this->determineModel($controller->prefix(), $statement->reference());
+                    $this->addImport($controller, $modelNamespace . '\\' . $model);
 
                     if ($statement->operation() === 'save') {
                         $tested_bits |= self::TESTS_SAVE;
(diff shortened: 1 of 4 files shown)


jasonmccreary:
This seems a bit heavy handed to _always_ add the import.

spekulatius:
That was my initial thought too. I looked around and found that all CRUD-methods would use the model, am I incorrect about this? I'm happy about any input in how to improve the PR.

jasonmccreary:
My initial thought is this is some kind of red-herring (non-issue). Are you sure you can reproduce the issue?

spekulatius:
Yeah, I've just reproduced it. Here are my steps:

```bash
cd /tmp
curl -s https://laravel.build/example-app | bash
cd example-app && composer install
composer require laravel-shift/blueprint
```

Then add the `draft.yaml`as mentioned in #583. 

```bash
php artisan blueprint:build
phpunit
```

leads to:

```
19:54 $ phpunit 
PHPUnit 9.5.27 by Sebastian Bergmann and contributors.

..E                                                                 3 / 3 (100%)

Time: 00:00.047, Memory: 22.00 MB

There was 1 error:

1) Tests\Feature\Http\Controllers\EntryControllerTest::show_displays_view
Error: Class "Tests\Feature\Http\Controllers\Entry" not found

/tmp/example-app/tests/Feature/Http/Controllers/EntryControllerTest.php:17
/home/peter/.config/composer/vendor/phpunit/phpunit/src/TextUI/Command.php:144
/home/peter/.config/composer/vendor/phpunit/phpunit/src/TextUI/Command.php:97

ERRORS!
Tests: 3, Assertions: 2, Errors: 1.
```

The generated file in question:

```php
<?php

namespace Tests\Feature\Http\Controllers;

use Tests\TestCase;

/**
 * @see \App\Http\Controllers\EntryController
 */
class EntryControllerTest extends TestCase
{
    /**
     * @test
     */
    public function show_displays_view()
    {
        $entry = Entry::factory()->create();

        $response = $this->get(route('entry.show', $entry));

        $response->assertOk();
        $response->assertViewIs('entry.show');
        $response->assertViewHas('entry');
    }
}
```

Maybe I've missed some steps, if so please let me know.

jasonmccreary:
What's happening is there is not `EloquentStatement` for the default _CRUD_ `show` action. It uses the route parameter directly for the view data.

As such, the code probably needs to be updated to ensure the model is imported for an `EloquentStatement` **and any models passed as route parameters**.

spekulatius:
Yeah, good point. I've checked and uncovered part in regards of model-import seems to be this place. It's what triggers my error by using the model without an import. I've tested it locally in the described setup and it fixed my issue.

jasonmccreary:
@spekulatius, great. It looks like all tests are passing, but I'd appreciate if you could retroactively add one to demonstrate this exact `show` issue.

spekulatius:
Cool, does [this cover](https://github.com/laravel-shift/blueprint/pull/584/commits/d549320bf5a7b0ecbc92af01957dfa94e908e6d9) the testing or does it need more tweaking?

jasonmccreary:
Looks good. I assume if you comment out the line you added, this test fails?

spekulatius:
At the moment it seems like it already fails on GitHub:

```
Mockery\Exception\InvalidCountException: Method put('tests/Feature/Http/Controllers/PostControllerTest.php', '<?php

namespace Tests\Feature\Http\Controllers;

// use App\Models\Post;
use Tests\TestCase;

/**
 * @see \App\Http\Controllers\PostController
 */
class PostControllerTest extends TestCase
{
    /**
     * @test
     */
    public function show_displays_view()
    {
        $post = Post::factory()->create();

        $response = $this->get(route('post.show', $post));

        $response->assertOk();
        $response->assertViewIs('post.show');
        $response->assertViewHas('post');
    }
}
') from Mockery_0_Illuminate_Filesystem_Filesystem should be called
 exactly 1 times but called 0 times.
```

The test files seem to be named correctly and placed correct with the other test files.

When running the test `draft.yaml` regularly in the project root the test is generated correctly. So I assume the yaml file is correct. Breaking it also leads to a different error when running the tests so it is considered. Any idea why the `put` isn't called during generation?

jasonmccreary:
This happens because what Blueprint is generating doesn't match the Mockery expectation. I wish they provided a "diff" instead of just the failure. You can add some debugging to see what Blueprint really outputs. It's often a whitespace or import order issue.

spekulatius:
Ah okay, I think I've got it now. This should pass the tests. I'll push the broken test after this one is confirmed to pass.

spekulatius:
Awesome, I've switched to `dev-master` and can confirm it's working as expected :+1: 

Thank you for checking it so close to Christmas :christmas_tree: Happy Holidays!

PR #586, opened 2022-12-23
Fix duplicate definitions in routes files

WeedTest:
Title: Fix duplicate definitions in routes files

Description:
This PR fixes the duplication in routes files
Changes Made:
- Added a alreadyAppended function to check if content already exits

Before:
        foreach (array_filter($routes) as $type => $definitions) {
            $path = 'routes/' . $type . '.php';
            $this->filesystem->append($path, $definitions . PHP_EOL);
            $paths[] = $path;
        }

        return ['updated' => $paths];

After:
        foreach (array_filter($routes) as $type => $definitions) {
            $path = 'routes/' . $type . '.php';
            if (!$this->alreadyAppended($path, $definitions)) {
                $this->filesystem->append($path, $definitions . PHP_EOL);
                $paths[] = $path;
            }
        }

        return empty($paths) ? [] : ['updated' => $paths];
....

    protected function alreadyAppended(string $path, string $definitions): bool
    {
        return str_contains($this->filesystem->get($path),$definitions);
    }


jasonmccreary:
I consider this a user error and not something Blueprint is responsible for or would ever get 100% correct. As such, I am closing this. Thanks though.

WeedTest:
ok i undestand

PR #591, opened 2023-01-05
Return correct return-type hints

### LINKED ISSUE #589: Wrong return type on controller, when returning a View
- Laravel Version: 9.38.0
- PHP Version: 8.1.12
- Blueprint Version: 2.4
- Platform: Mac

### Issue:

I set `'use_return_types' => true,` in `config/blueprint.php`. Blueprint adds the return types but does not take the real return type into account, which in case of a `create()` action is `View`.

```php
class UserController extends Controller
{
    /**
     * @param \Illuminate\Http\Request $request
     * @return \Illuminate\Http\Response
     */
    public function create(Request $request): \Illuminate\Http\Response
    {
        return view('user.create');
    }
   
    //...    
}
```

### draft.yaml:
```yaml
models:
  User:
    name: string
    email: string unique
    email_verified_at: datetime nullable
    password: string
    remember_token: rememberToken
    softDeletes

controllers:
  User:
    resource
```

ghostwriter:
Resolves #589


jasonmccreary:
You can run `vendor/bin/pint` locally to fix everything you need.

Issue #611, opened 2023-05-04
Support for multiple `send` key items for controllers


faustbrian:
### Synopsis:

Allow specifying multiple items for the `send` key to send multiple notifications.

### Proposed Syntax:

```yml
Ticket:
  resource: all
  store:
    validate: title, content
    save: ticket
    send:
      - TicketReceived to:ticket.customer with:ticket
      - NewTicket to:ticket.staff with:ticket
    dispatch: SyncMedia with:ticket
    fire: NewTicket with:ticket
    flash: ticket.id
    redirect: ticket.index
```

### Expected Behavior:

The proposed syntax should generate multiple lines.

```php
Notification::send($ticket->customer, new TicketReceived($ticket));
Notification::send($ticket->staff, new NewTicket($ticket));
```

faustbrian:
That currently fails since duplicate keys are not supported in YAML nor JSON.

```
php artisan blueprint:build

   Symfony\Component\Yaml\Exception\ParseException

  Duplicate key "send" detected at line 538 (near "send: A to:ticket.author with:ticket").

  at vendor/symfony/yaml/Parser.php:348
    344▕                     // But overwriting is allowed when a merge node is used in current block.
    345▕                     if ($allowOverwrite || !isset($data[$key])) {
    346▕                         $data[$key] = $value;
    347▕                     } else {
  ➜ 348▕                         throw new ParseException(sprintf('Duplicate key "%s" detected.', $key), $this->getRealCurrentLineNb() + 1, $this->currentLine);
    349▕                     }
    350▕                 }
    351▕                 if ($isRef) {
    352▕                     $this->refs[$isRef] = $data[$key];
```

jasonmccreary:
Ah, lame YAML. I have reopened this for the community to implement.

faustbrian:
Will take a look in the next days and see if I can figure out where to get the additional parsing working 👀

jasonmccreary:
I'd probably look at the Statement "lexer". I believe there is where you could check and handle if the value is an array, if so, register multiple `SendStatement`. The rest should just work.

faustbrian:
Just had a cursory look and due to how the parser seems to work it results in something like this:

```php
      "store" => array:7 [
        "validate" => "title, content"
        "save" => "ticket"
        "send" => "ReviewNotification to:ticket.author with:ticket ReviewNotification to:ticket.author with:ticket"
        "dispatch" => "SyncMedia with:ticket"
[...]

PR #612, opened 2023-05-05
Allow multiple send statement values for controllers

faustbrian:
> Keeping this as draft until https://github.com/laravel-shift/blueprint/pull/619 is merged.

Resolves https://github.com/laravel-shift/blueprint/issues/611 by making as little changes as possible in the form of using the `[]` syntax for arrays and using `Arr::wrap` in various places to add support for handling the `send` array but this also means other properties can now use arrays as their statement value in the future and it should be handled in some places.

**Input**
```yml
controllers:
  Ticket:
    store:
      validate: title, content
      save: ticket
      send: [
        ReviewNotification to:ticket.author with:ticket,
        ReviewNotification to:ticket.author with:ticket
      ]
      dispatch: SyncMedia with:ticket
      fire: NewTicket with:ticket
      flash: ticket.id
      redirect: ticket.index
```

**Output**

```php
<?php

namespace App\Http\Controllers;

use App\Events\NewTicket;
use App\Http\Requests\TicketStoreRequest;
use App\Http\Requests\TicketUpdateRequest;
use App\Jobs\SyncMedia;
use App\Models\Ticket;
use App\Notification\ReviewNotification;
use Illuminate\Http\RedirectResponse;
use Illuminate\Http\Request;
use Illuminate\Support\Facades\Notification;
use Illuminate\View\View;

class TicketController extends Controller
{
    public function store(TicketStoreRequest $request): RedirectResponse
    {
        $ticket = Ticket::create($request->validated());

        Notification::send($ticket->author, new ReviewNotification($ticket));
        Notification::send($ticket->author, new ReviewNotification($ticket));

        SyncMedia::dispatch($ticket);

        event(new NewTicket($ticket));

        $request->session()->flash('ticket.id', $ticket->id);

        return redirect()->route('ticket.index');
    }
}
```

### DIFF
--- a/src/Generators/ControllerGenerator.php
+++ b/src/Generators/ControllerGenerator.php
@@ -18,6 +18,7 @@ use Blueprint\Models\Statements\SendStatement;
 use Blueprint\Models\Statements\SessionStatement;
 use Blueprint\Models\Statements\ValidateStatement;
 use Blueprint\Tree;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 
 class ControllerGenerator extends AbstractClassGenerator implements Generator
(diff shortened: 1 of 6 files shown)


jasonmccreary:
There must be a way, without brackets or dashes, to support sub statements as we do something similar for model relationships. Even the very syntax in the draft file simply uses indents for statements. So there is some mechanism which should support:

```yaml
      send: 
        ReviewNotification to:ticket.author with:ticket
        ReviewNotification to:ticket.author with:ticket
```

faustbrian:
Relationships are not an array since they have keys in the form of `relationshipType: modelList` which means they are an object instead of an array.

jasonmccreary:
I'm not completely against the array notation. However, it introduces a rather bit of dev syntax to an otherwise readable YAML file. As such, I'm not ready to give up on something simpler.

Two (quick) ideas:
- Allow for multiple `send` statements, but pre-parse the YAML (like we do with shorthands) to suffix them, i.e. `send-1`
- Key the sub statements so it's an object like model relationships, i.e. `ReviewNotification: to:post.author with:post`

faustbrian:
I'll take a look at the first option, might have an idea how to solve it without requiring the numeric suffixes.

PR #619, opened 2023-05-05
Add support for iterating over statement children

faustbrian:
> Extracted from https://github.com/laravel-shift/blueprint/pull/612 to keep the PRs easier to digest and this is more of a prerequisite for the mentioned PR rather than a part of it. 

This PR makes it possible to handle statements where the value is an array, as will be the case after merging the above mentioned PR.


jasonmccreary:
I'm going to merge this. But before any downstream PRs are merged, I want to refactor this.

Issue #340, opened 2020-08-21
Friendly warning for missing models


jasonmccreary:
### Synopsis:
It's possible if a user runs `blueprint:build` for an existing app (or a `draft.yaml` file with references the `User` model) it's likely they may receive a nasty error.

This is often mitigated by running `blueprint:trace`. Yet there is no indication to do so. In the end, this is a poor out-of-the-box experience. 

### Expected Behavior:
Given the following draft.yaml file, it would be better if an error message were thrown letting the user know they have an "unknown model" and to either define it or run `blueprint:trace` if it exists already.

```yaml
controllers:
  User:
    resource: api
```

A centralized place to throw this error might be `modelForContext`. However, sometimes Blueprint handles a missing models. So these will likely need to be reviewed, but maybe a second `throw` parameter could be added.

devmsh:
I will give it a try 👍

devmsh:
While `modelForContext` invoked in the generation phase, I think we can check the missing models earlier after the analysis phase even before generate any valid controllers and throw the exception in the middle of the process.

What about adding extra validation step inside `Blueprint::analyze` where we have the full `Tree` and it may be extended in the future to make extra validations?

PR #622, opened 2023-05-05
Throw an exception when a required model is missing

faustbrian:
> Resolves https://github.com/laravel-shift/blueprint/issues/340

Currently it only throws in two places, based on the cases I could track down where `modelForContext` is called and afterwards a `Model` instance is expected or unhandled exceptions will be encountered.


jasonmccreary:
What is a place where we wouldn't throw an exception?

PR #623, opened 2023-05-05
Allow multiple dispatch, fire, notify and send statements

faustbrian:
> Alternative to https://github.com/laravel-shift/blueprint/pull/612 for https://github.com/laravel-shift/blueprint/issues/611

This new approach preprocesses the YAML content and transforms duplicate keys into keys that are suffixed with their line number to avoid key collisions. This also reverts https://github.com/laravel-shift/blueprint/pull/619.

**Input**
https://github.com/laravel-shift/blueprint/pull/623/files#diff-a4f4e06b3fdb4a8424880f544fca3c75a0ecf2ebb5158feba4963f7066e2d265

**Output**
https://github.com/laravel-shift/blueprint/pull/623/files#diff-72d9e08bf23239b5487ea8767fe78e8f3e9be1b14e5bd64e1ae2c4848da167f1

### DIFF
--- a/src/Blueprint.php
+++ b/src/Blueprint.php
@@ -38,6 +38,8 @@ class Blueprint
             $content = preg_replace('/^(\s*)-\s*/m', '\1', $content);
         }
 
+        $content = $this->transformDuplicatePropertyKeys($content, ['dispatch', 'fire', 'notify', 'send']);
+
         $content = preg_replace_callback(
             '/^(\s+)(id|timestamps(Tz)?|softDeletes(Tz)?)$/mi',
             fn ($matches) => $matches[1] . strtolower($matches[2]) . ': ' . $matches[2],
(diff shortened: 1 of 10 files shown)


faustbrian:
This requires quite some regex voodoo to get it working. It's almost working but there seems to be something somewhere that causes duplication, for example when you run `vendor/bin/phpunit --filter="output_writes_events"` you'll get duplicated events. Not sure yet what causes that but I assume the multiple calls to the controller lexer.

jasonmccreary:
I can take a stab if you want to leave it for the weekend.

faustbrian:
I'll continue tomorrow on it but feel free to poke around, probably something obvious that causes the array items to get duplicated 😅

jasonmccreary:
Cool. You on Twitter? I'll give you a shout out.

jasonmccreary:
I think parsing out the `controller` section via regex will prove challenging and likely have edge cases. If you wanted to produce the dashed array syntax, then you'd might be better off going line by line through the controller section.

However, thinking about it more, these statements may not be sequential. And moving them into a single array, may not generate the code the user intended.

For example:

```yaml
      validate: title, content
      save: ticket
      send: ReviewNotification to:ticket.reviewer with:ticket
      dispatch: SyncMedia with:ticket
      send: TicketOpened to:ticket.author with:ticket
```

As such, I think simply finding these indented properties and giving them a numerical suffix is the way to go. Then update the lexer to handle numbered properties when creating the statement. If so, you could likely revert #612.

src/Blueprint.php
```diff
@@ -136,4 +138,68 @@ protected function shouldGenerate(array $types, array $only, array $skip): bool
 
         return true;
     }
+
+    private function transformPropertiesToArray(string $content, array $properties): string
+    {
+        preg_match_all('/controllers:(?:\n|.)*$/m', $content, $controllers);
```
jasonmccreary, line 144: Simply find indented properties at the beginning of the line and add a numeric suffix.

src/Lexers/StatementLexer.php
```diff
@@ -14,6 +14,7 @@
 use Blueprint\Models\Statements\SendStatement;
 use Blueprint\Models\Statements\SessionStatement;
 use Blueprint\Models\Statements\ValidateStatement;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 
 class StatementLexer implements Lexer
```
jasonmccreary, line 20: In `analyze` simply allow matching of `send-*`, `fire-*`, etc.

faustbrian: What's your preferred way of handling that? Simply `'fire', 'fire-*' => $this->analyzeEvent($statement)` obviously won't work since a `match` statement on a literal string is used.

faustbrian: Went with `fnmatch` since that allows easy glob matching.

faustbrian:
@jasonmccreary updated the OP with the (I think) final changes

faustbrian:
Not sure why windows is failing 🤔

faustbrian:
@jasonmccreary do you happen to have a windows machine to play with the data and see why it fails with the new keys?

jasonmccreary:
I foresee a lot of bugs, but here we go...

Issue #618, opened 2023-05-05
Support for constructor property promotion


faustbrian:
### Synopsis:

Following https://github.com/laravel-shift/blueprint/issues/617 it would be useful to make constructor property promotion configurable. This would allow for code generation that requires less refactoring after the fact. Even more fine-grained control would be to allow the use of `readonly` to further reduce the need for refactoring if someone wishes to use `readonly` in combination with constructor property promotion.

### Proposed Syntax:

A simple boolean flag in the configuration that indicates whether or not templates should use constructor property promotion.

### Expected Behavior:

```php
<?php

namespace App\Events;

use App\Models\Ticket;
use Illuminate\Queue\SerializesModels;

class NewTicket
{
    use SerializesModels;

    /**
     * Create a new event instance.
     */
    public function __construct(public Ticket $ticket)
    {
        //
    }
}
```

PR #624, opened 2023-05-06
Add support for constructor property promotion

faustbrian:
Resolves https://github.com/laravel-shift/blueprint/issues/618


jasonmccreary:
I'll merge this since it was requested and added by the community. But not something I would do personally.

PR #642, opened 2023-06-18
FactoryGenerator: No need to import fqcn foreign keys.

Jehong-Ahn:
No need `use App\Models\\Foo\Bar;` for `foreign:\Foo\Bar`.

### DIFF
--- a/src/Generators/FactoryGenerator.php
+++ b/src/Generators/FactoryGenerator.php
@@ -97,9 +97,10 @@ class FactoryGenerator extends AbstractClassGenerator implements Generator
                 }
 
                 $class = Str::studly(Str::singular($table));
-                $reference = $this->fullyQualifyModelReference($class) ?? $model;
-
-                $this->addImport($model, $reference->fullyQualifiedNamespace() . '\\' . $class);
+                if (!Str::startsWith($class, '\\')) {
+                    $reference = $this->fullyQualifyModelReference($class) ?? $model;
+                    $this->addImport($model, $reference->fullyQualifiedNamespace() . '\\' . $class);
+                }
 
                 if ($key === 'id') {
                     $definition .= str_repeat(self::INDENT, 3) . "'{$column->name()}' => ";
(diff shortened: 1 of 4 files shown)


src/Generators/FactoryGenerator.php
```diff
@@ -97,9 +97,10 @@ protected function buildDefinition(Model $model)
                 }
 
                 $class = Str::studly(Str::singular($table));
-                $reference = $this->fullyQualifyModelReference($class) ?? $model;
-
-                $this->addImport($model, $reference->fullyQualifiedNamespace() . '\\' . $class);
+                if (!Str::startsWith($class, '\\')) {
+                    $reference = $this->fullyQualifyModelReference($class) ?? $model;
+                    $this->addImport($model, $reference->fullyQualifiedNamespace() . '\\' . $class);
+                }
```
ghostwriter, line 103: This is wrong.

---

I believe you want to remove the application's namespace prefix for fully-qualified class names that start with '\\'.

eg. for `foreign:\Foo\Bar`

```diff
- use App\Models\\Foo\Bar;
+ use Foo\Bar;
```

To achieve this we need to change:

```diff
$class = Str::studly(Str::singular($table));
$reference = $this->fullyQualifyModelReference($class) ?? $model;

- $this->addImport($model, $reference->fullyQualifiedNamespace() . '\\' . $class);
+ if (Str::startsWith($class, '\\')) {
+     $this->addImport($model, $class);
+ } else {
+     $this->addImport($model, $reference->fullyQualifiedNamespace() . '\\' . $class);
+ }
```
---

This should correctly import **all** fully-qualified class names.

ghostwriter (changes requested):
Hey @Jehong-Ahn ,

We appreciate the time you invested in preparing this pull request.

I will be reviewing this patch now.

Thank you for your contribution.

jasonmccreary:
What's the status of this PR. Seems conversations are not resolved.

PR #643, opened 2023-06-22
use table name for belongsToMany()

Jehong-Ahn:
belongsToMany() needs a table name always.

```php
if ($is_pivot) {
    ...
    $relationship = sprintf('$this->%s(%s::class, \'%s\')', $type, $fqcn, $foreign->tableName());
    ...
}
else {
    $relationship = sprintf('$this->%s(%s::class, \'%s\')', $type, $fqcn, $column_name);
}
```

### DIFF
--- a/src/Generators/ModelGenerator.php
+++ b/src/Generators/ModelGenerator.php
@@ -225,11 +225,12 @@ class ModelGenerator extends AbstractClassGenerator implements Generator
                     $relationship = sprintf('$this->%s(%s::class, \'%s\', \'%s\')', $type, $fqcn, $column_name, $key);
                 } elseif (!is_null($class) && $type === 'belongsToMany') {
                     if ($is_pivot) {
-                        $relationship = sprintf('$this->%s(%s::class)', $type, $fqcn);
+                        $foreign = $this->tree->modelForContext($column_name, true);
+
+                        $relationship = sprintf('$this->%s(%s::class, \'%s\')', $type, $fqcn, $foreign->tableName());
                         $relationship .= sprintf('%s->using(%s::class)', PHP_EOL . str_pad(' ', 12), $column_name);
                         $relationship .= sprintf('%s->as(\'%s\')', PHP_EOL . str_pad(' ', 12), Str::snake($column_name));
 
-                        $foreign = $this->tree->modelForContext($column_name, true);
                         $columns = $this->pivotColumns($foreign->columns(), $foreign->relationships());
                         if ($columns) {
                             $relationship .= sprintf('%s->withPivot(\'%s\')', PHP_EOL . str_pad(' ', 12), implode("', '", $columns));
--- a/tests/fixtures/drafts/custom-pivot.yaml
+++ b/tests/fixtures/drafts/custom-pivot.yaml
@@ -12,6 +12,6 @@ models:
   Membership:
     meta:
       pivot: true
-      table: team_user
+      table: another_table_name
     user_id: id
     team_id: id
--- a/tests/fixtures/models/custom-pivot-membership.php
+++ b/tests/fixtures/models/custom-pivot-membership.php
@@ -15,7 +15,7 @@ class Membership extends Pivot
      *
      * @var string
      */
-    protected $table = 'team_user';
+    protected $table = 'another_table_name';
 
     /**
      * Indicates if the IDs are auto-incrementing.
--- a/tests/fixtures/models/custom-pivot-team.php
+++ b/tests/fixtures/models/custom-pivot-team.php
@@ -30,7 +30,7 @@ class Team extends Model
 
     public function users(): BelongsToMany
     {
-        return $this->belongsToMany(User::class)
+        return $this->belongsToMany(User::class, 'another_table_name')
             ->using(Membership::class)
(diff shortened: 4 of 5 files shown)


jasonmccreary:
I don't think it's true that it **always** needs a name. Only when it's unconventional. In addition, all tests are failing.

Jehong-Ahn:
I know table name convention. But, how can I use unconventional pivot table name?
For non-pivot belongsToMany(), the table name is already being supplied.
`$relationship = sprintf('$this->%s(%s::class, \'%s\')', $type, $fqcn, $column_name);`

I didn't test the whole case. My mistake.
Here is the missing commit.
https://github.com/laravel-shift/blueprint/compare/master...Jehong-Ahn:blueprint:belongsToMany-tablename

But why close so quickly? Always do you?

ghostwriter:
Hey @Jehong-Ahn,

Don't take offense, it's not personal. 

Closing issues/pull requests early is normal practice that helps maintainers focus on bugs and features that require attention by only leaving those issues open.

Plus we can always reopen any issue/pull request for further discussion.

Now on to reviewing this patch.

---

I've made the same mistake before, don't change existing test fixtures unless you are fixing a bug.

Instead, create a new draft file in the fixtures directory.

---

I don't think this pull request is necessary because, correct me if I'm wrong, Laravel automatically uses the value of `$table` on the Pivot model to determine the table name which can be set via `meta` configuration.

```
Membership:
    meta:
      pivot: true
      table:custom_table_name
    user_id: id
    team_id: id
```

```php
class Membership extends Pivot
{
    protected $table = 'custom_table_name';
}

class User extends Model
{
    public function teams(): BelongsToMany
    {
        return $this->belongsToMany(Team::class) // <-- no need to add Pivot table name
            ->using(Membership::class) // <- because table name is derived from `Membership::$table`
            ->as('membership')
            ->withPivot('id')
            ->withTimestamps();
    }
}

class Team extends Model
{
    public function users(): BelongsToMany
    {
        return $this->belongsToMany(User::class) // <-- no need to add Pivot table name
            ->using(Membership::class) // <- because table name is derived from `Membership::$table`
            ->as('membership')
            ->withPivot('id')
            ->withTimestamps();
    }
}
```

Thanks again for your contributions.

Jehong-Ahn:
Thanks, @ghostwriter 

Unfortunately, laravel don't use table name property. 
On belongsToMany(), laravel forces conventional table name.

Maybe this issue is on the laravel, not blueprint.

https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php#L513
```php
        // If no table name was provided, we can guess it by concatenating the two
        // models using underscores in alphabetical order. The two model names
        // are transformed to snake case from their default CamelCase also.
        if (is_null($table)) {
            $table = $this->joiningTable($related, $instance);
        }
```

Jehong-Ahn:
https://github.com/laravel/framework/issues/29571

ghostwriter:
No, it's not on Laravel's side. 

You're right, I believe I understand what you're looking for.

It's seems to me like we need to implement something that tells blueprint to use custom table name that's passed to `meta.table` configuration.

Only If 

1. `meta.table` was set to a non-empty-string in draft file
2. `meta.table` is `!==` the convention `snake cased models sorted alphabetically and concatenated with an underscore`
3. `meta.pivot` is set to true.

Adding a test that explicitly tests these requirement should get the rest of the tests passing and achieve what you need.

ghostwriter:
My previous answer is based on the recommendations https://github.com/laravel/framework/issues/29571#issuecomment-521333335

It's definitely worth discussing with Laravel internals to see if they will accept a pull request to make sure Laravel starts using the   `$table` property on the pivot models.

Jehong-Ahn:
https://github.com/laravel/framework/issues/21652

They say, this isn't a bug. It's weird.

jasonmccreary:
PRs that don't have the correct change, tests, or contain a feature I don't wish to support within Blueprint get closed. That's how it works. If you want, you are welcome to fork this package and maintain your own version. 👍 

If there is something you feel should still be added, please leave a clear comment demonstrating the problem.

Jehong-Ahn:
OK. Thanks for commenting. 👍

dansleboby:
> Thanks, @ghostwriter
> 
> Unfortunately, laravel doesn't use table name property for pivot relation. On belongsToMany(), laravel forces conventional table name.
> 
> Maybe this issue is on the laravel, not blueprint.
> 
> https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php#L513
> 
> ```
>         // If no table name was provided, we can guess it by concatenating the two
>         // models using underscores in alphabetical order. The two model names
>         // are transformed to snake case from their default CamelCase also.
>         if (is_null($table)) {
>             $table = $this->joiningTable($related, $instance);
>         }
> ```

Hi if you use something like this it will take the table name in the pivot class

```
$this->belongsToMany(Team::class, Membership::class) // <-- no need to add Pivot table name
            ->using(Membership::class)
```

Jehong-Ahn:
Thank you @dansleboby

PR #649, opened 2023-11-20
Fix #648 string interpolation for php8.2

### LINKED ISSUE #648: String interpolation ${} in PHP8.2 warning
- Laravel Version: 10.32.1
- PHP Version: 8.2.12
- Blueprint Version: 2.7.0
- Platform: Mac (actually Linux on Docker, Laravel Sail)

### Issue:
Running `blueprint:new` (and probably other blueprint commands) in PHP 8.2 gives this warning. The yaml file was created, but I switched to 8.1 so I didn't have to deal with it.

```
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in /var/www/html/vendor/laravel-shift/blueprint/src/Generators/TestGenerator.php on line 525
```

~inpeto

impeto:
All tests are green on php8.1 and 8.2. Let me know.

Thanks,

~impeto

### DIFF
--- a/src/Generators/PestTestGenerator.php
+++ b/src/Generators/PestTestGenerator.php
@@ -538,14 +538,15 @@ class PestTestGenerator extends AbstractClassGenerator implements Generator
 
     private function buildFormRequestTestCase(string $controller, string $action, string $form_request)
     {
-        return <<< END
-test('${action} uses form request validation')
+        $template =<<< END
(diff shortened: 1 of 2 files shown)


jasonmccreary:
Thanks, but the `sprintf` is overkill. Just need to move the `$` inside the braces.

PR #657, opened 2023-12-09
Add Support for ULID column type

rcrosbourne:
List of changes:

1. Refactor foreign key generation in migrations
2. Refactored the foreign key generation code in migration files into a method.
3. Supports generation of primary keys with 'ulid' and 'uuid' data types for pivot tables.
4. Added support for ULID and UUID in pivot tables migration
5. Expanded the migration generation function to support ULID and UUID primary keys.
6. Add ULID support in `Model` and `ModelGenerator` classes
7. Function to check if ULIDs are used in the `Model` class.
8. Expand 'ModelGenerator' class to act accordingly if ULIDs are used.
9. Add support for nullableUlidMorphs and ulidMorphs column types
10. The column types have been added to the all-column-types test fixtures and model generator.
11. Update Foreign Key Constraints and Migration Generator
12. Foreign key constraints updated in model test fixtures and new foreign keys `customer_id` and `tran_id` added.
13. 14. `ulidMorphs` added to the test fixtures in all-column-types for drafts, models, and factories.
14. Add test for ULID without relationship scenario
15. A new test for the case of a ULID without a relationship was added in `MigrationGeneratorTest.php` file.
16. Add ULID support to FactoryGenerator and tests
17. Add support for ULID shorthand in migrations
18. Add `ulid` type to `ModelLexer.php`.
19. Update regular expression in `Blueprint.php` to match `ulid` and `uuid`.


jasonmccreary:
Sorry, this commit is way too large with too many commits and changes for me to consider. If you want a new feature added, I recommend separating your PR into multiple. First making the feature, then additional for refactors.

rcrosbourne:
Thanks for your feedback. This feature involved adding`ulid` support for models, migrations and factories. So there would be a lot of files changes. I can take a phased approach by adding support for migrations only and slowly add the rest.

tests/Feature/BlueprintTest.php
```diff
@@ -117,6 +117,22 @@ public function it_parses_shorthands(): void
...
 
+    #[Test]
+    public function it_parses_ulid_shorthand(): void
+    {
+        $blueprint = $this->fixture('drafts/ulid-shorthand.yaml');
+
+        $this->assertEquals([
+            'models' => [
+                'Person' => [
+                    'id' => 'ulid primary',
```
jasonmccreary, line 128: It should not be required to say `ulid primary`, just `ulid` is fine as it is the conventional `id` field.

src/Generators/MigrationGenerator.php
```diff
@@ -128,10 +128,10 @@ protected function populateStub(string $stub, Model $model): string
         return $stub;
     }
 
-    protected function populatePivotStub(string $stub, array $segments): string
+    protected function populatePivotStub(string $stub, array $segments, array $models = []): string
```
jasonmccreary, line 131: Why is `models` needed?

src/Generators/MigrationGenerator.php
```diff
@@ -194,7 +194,7 @@ protected function buildDefinition(Model $model): string
                 );
             }
 
-            if (!empty($columnAttributes) && !$this->isIdOrUuid($column->dataType())) {
+            if (!empty($columnAttributes) && !$this->isIdUlidOrUuid($column->dataType())) {
```
jasonmccreary, line 197: This method is getting a bit silly, please refactor it to `isIdColumnType`

src/Generators/MigrationGenerator.php
```diff
@@ -304,13 +303,49 @@ protected function buildPivotTableDefinition(array $segments): string
...
+     * to the `$definition` string. If the model does not exist or its name does not match the
+     * pivot table segment, it defaults to appending `$table->foreignId(\'' . $foreignKeyColumnName . '\');`
+     * to the `$definition` string. The function then returns the `$definition` string.
+     *
+     * @param  string  $pivotTableSegment The segment of the pivot table. e.g 'dive_job' it would be 'Dive' or 'Job'.
+     * @param  string  $foreignKeyColumnName The name of the foreign key column. e.g 'dive_id' or 'job_id'.
+     * @param  array  $models An array of models. e.g ['Dive' => $diveModel, 'Job' => $jobModel].
+     * @return string The foreign key definition. e.g '$table->foreignUlid('dive_id');'
+     */
+    protected function generateForeignKeyDefinition(string $pivotTableSegment, string $foreignKeyColumnName, array $models = []): string
```
jasonmccreary, line 328: Not sure I understand the reason for a `foreach` here. The original code was a single line, why do we need to look now?

jasonmccreary (commented):
Okay, I gave it a quick review. Please fix the comments and run `vendor/bin/pint` locally.

src/Generators/MigrationGenerator.php
rcrosbourne: So the issue here is when two models have a Many to Many relationship, the pivot table was defaulting to `foreignId` which isn't correct if the Models has a `uuid` or `ulid` column type. This function checks the model `id` type and if its ulid it returns `foreignUlid`, `uuid` it returns `foreignUuid` and `foreignId` otherwise. 
E.g
Before 
```yaml
models:
  Diver:
    id: ulid primary
    name: string
    base_pay: decimal:10,2
    relationships:
      belongsToMany: Job
  Job:
    id: ulid primary
    location: string
    vessel: string
    relationships:
      belongsToMany: Diver
```
Would yield the pivot migration
```php
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

return new class extends Migration
{
    /**
     * Run the migrations.
     */
    public function up(): void
    {
        Schema::create('diver_job', function (Blueprint $table) {
            $table->foreignId('diver_id');
            $table->foreignId('job_id');
        });
    }

    /**
     * Reverse the migrations.
     */
    public function down(): void
    {
        Schema::dropIfExists('diver_job');
    }
};
```
After this change
```php
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

return new class extends Migration
{
    /**
     * Run the migrations.
     */
    public function up(): void
    {
        Schema::create('diver_job', function (Blueprint $table) {
            $table->foreignUlid('diver_id');
            $table->foreignUlid('job_id');
        });
    }

    /**
     * Reverse the migrations.
     */
    public function down(): void
    {
        Schema::dropIfExists('diver_job');
    }
};
```

rcrosbourne: There is probably a cleaner way to do this. I will defer to you.

tests/Feature/BlueprintTest.php
rcrosbourne: It isn't required to specify `ulid primary`. Here is the [draft](https://github.com/laravel-shift/blueprint/pull/657/files#diff-20e2cd8edfa713aaf638d0916fbda107837b16fd178030ca9c8b936a04ef4283) file that was used in that test `tests/fixtures/drafts/ulid-shorthand.yaml`

src/Generators/MigrationGenerator.php
rcrosbourne: Agreed :)

src/Generators/MigrationGenerator.php
rcrosbourne: The original [code](https://github.com/laravel-shift/blueprint/blob/master/src/Generators/MigrationGenerator.php#L297) had the `foreach`. This function replaces this [line](https://github.com/laravel-shift/blueprint/blob/master/src/Generators/MigrationGenerator.php#L307) in the original function

src/Generators/MigrationGenerator.php
jasonmccreary: I see. Thanks.

src/Generators/MigrationGenerator.php
jasonmccreary: Ok. Please change it.

tests/Feature/BlueprintTest.php
jasonmccreary: Since it is not needed, please remove `ulid primary` from all of your new fixtures to avoid confusion for those reviewing them as examples.

src/Generators/MigrationGenerator.php
jasonmccreary: I think this has to do with the pivot table needing to review other models for their `id` column type.

jasonmccreary:
Please resolve any open conversation and I'll give this a final review.

tests/Feature/BlueprintTest.php
rcrosbourne: Fixed.

jasonmccreary:
Looks like there's still a bit of lint. Run `composer update` locally and then `vendor/bin/pint`. It's probably a new rule in Pint that's not being applied.

tests/Feature/BlueprintTest.php
rcrosbourne: Fixed.

rcrosbourne:
> Looks like there's still a bit of lint. Run `composer update` locally and then `vendor/bin/pint`. It's probably a new rule in Pint that's not being applied.

Yes that worked. Thanks.

Issue #662, opened 2024-01-18
Plural route naming convention as default to match Laravel docs


bertheyman:
### Synopsis:
[This debate](https://github.com/laravel-shift/blueprint/issues/480) and [subsequent PR](https://github.com/laravel-shift/blueprint/pull/574/files) allowed for routes names to be configurable: singular or plural.

The Blueprint default differs from the Laravel docs:

Blueprint
```php
Route::resource('photo', PhotoController::class);
```

Laravel docs
```php
Route::resource('photos', PhotoController::class);
```

### Proposed Syntax:

```php

// config/blueprint.php

'singular_routes' => null,
```

### Expected Behavior:

Have plural as the default option to match the docs and promote Laravel conventions.
I realise there is no 100% official naming conventions, but I believe sticking to the docs will help promote one standard.

jasonmccreary:
I agree that the plural/singular is a bit off. If you want to open a PR to correct some of this, I am glad to revisit it.

PR #663, opened 2024-02-01
Plural routes by default

bertheyman:
This uses the plural of a modal for resource controllers by default to match the examples in the Laravel docs as discussed in #662 . Singular still remains possible via a config setting.

I checked the docs and could not find any references to the singular ones, so no need for changes there as far as I could see.

### DIFF
--- a/config/blueprint.php
+++ b/config/blueprint.php
@@ -106,15 +106,15 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Pluralize route names
+    | Singular route names
     |--------------------------------------------------------------------------
     |
     | By default, Blueprint will use the `kebab-case` of the controller name
-    | for the route name. If you would like to ensure a plural route name
+    | for the route name. If you would like to ensure a singular route name
     | is used, you may set this to `true`.
     |
     */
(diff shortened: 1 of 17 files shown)


jasonmccreary:
This is great. Please fix the linting errors.

Also, what about the views path? Again, probably not something documented, but I would imagine they should be plural by default too?

bertheyman:
Thanks for checking!
I've fixed the linting error.

For the views path, the most official I could find are some examples from the docs.

Most use stuff like 'welcome.blade.php', but anything model related does use singular.

Example from the [docs](https://laravel.com/docs/10.x/routing#implicit-binding):
```php
// Route definition...
Route::get('/users/{user}', [UserController::class, 'show']);
 
// Controller method definition...
public function show(User $user)
{
    return view('user.profile', ['user' => $user]);
}
```

The [mailing section](https://laravel.com/docs/10.x/mail#configuring-the-view) also uses singular.

**TL;DR: current approach of Blueprint matches view folder conventions from the docs**.

jasonmccreary:
Fair enough. Thanks for this!

Issue #658, opened 2023-12-10
Change generator methods visibility from "private" to "protected"


mowses:
### Synopsis:
I would like to customize the Blueprint generators a bit (for example `/src/Generators/ModelGenerator.php`, so instead of copying the original class and modify it, I would like to just extend from the original class and override the methods I need.
This way I keep my implementation slim and still can get the updates from the original file.

The problem I faced was with the private methods, which I could not override.
I tried to find a protected parent method to override and create my own implementation, but in the end I came up with almost the entire class replicated, just to customize a single method.

I would like to suggest to change from private methods to protected, since it can be extended when needed.

### Proposed Syntax:
Change private methods to protected.

### Expected Behavior:
It wont break anything and still can be extended and customized from the main application.

jasonmccreary:
Feel free to open a PR with the methods you need `protected`. I ask that you do it consistently to all _generators_.

jasonmccreary:
I have reviewed all of the Generators and opened up any method I felt might make sense to override. What's left is internal or dictated by Laravel convention. If someone needs access to that method, they can override the calling method.

PR #666, opened 2024-02-13
Protected methods for generators

bertheyman:
This PR changes the visibility of generator methods to at least protected to allow for easier customisation.
Based on the request in [this issue](https://github.com/laravel-shift/blueprint/issues/658).


jasonmccreary:
My gut feeling is this will cause more headaches. As noted in the open issue, if there is a particular method that needs to be "open", I'm glad to consider it. But opening all methods of all generators seems extreme.

PR #668, opened 2024-02-16
Remove DBAL for Laravel 11

jasonmccreary:


jasonmccreary:
While the tests are passing it seems some of the column attributes don't match up when testing the `README` example. I'm not sure if this is due to Laravel 10 versus Laravel 11, SQLite versus MySQL, or DBAL versus Laravel.

hafezdivandari:
@jasonmccreary it's pretty good, let me send a PR based on your branch, then you can see if it helps.

hafezdivandari:
@jasonmccreary you may check this #669

PR #677, opened 2024-02-29
Fix plural route defaults

jasonmccreary:
This corrects some of the areas missed in #663 to make routes plural by default.

- [x] Tests generate code for plural/singular routes
- [x] Resource shorthands generate plural routes
- [x] Update draft fixtures to use plural routes
- [ ] Update `README.md` example


src/Generators/PestTestGenerator.php
```diff
@@ -376,10 +376,7 @@ protected function buildTestCases(Controller $controller): string
                 } elseif ($statement instanceof RedirectStatement) {
                     $tested_bits |= self::TESTS_REDIRECT;
 
-                    $assertion = sprintf(
-                        '$response->assertRedirect(route(\'%s\'',
-                        config('blueprint.plural_routes') ? Str::plural(Str::kebab($statement->route())) : Str::kebab($statement->route())
-                    );
+                    $assertion = sprintf('$response->assertRedirect(route(\'%s\'', $statement->route());
```
jasonmccreary, line 380: Whatever route the user specifies in their _draft file_ should be used directly.

PR #692, opened 2024-06-19
Make Pest the default testing option

bertheyman:
Since Laravel 11, Pest is the default testing option but the developer can opt to use PHPUnit.
Would it make sense to do the same in Blueprint?

If so, this PR makes Pest the default testing option.


jasonmccreary:
To my knowledge, [Pest is not the default](https://github.com/laravel/laravel/blob/11.x/composer.json). Maybe you created a new app with the option for Pest.

Either way, it's easy to [swap out in Blueprint's config](https://blueprint.laravelshift.com/docs/generated-tests/).

bertheyman:
I based myself on [this article](https://laraveldaily.com/post/laravel-11-main-new-features-changes) (new defaults: Pest), the [Tweet from Taylor](https://x.com/taylorotwell/status/1744729110163988949) and Pest being shown first in the examples.

Reading more about it: there still seems to be a lot of debate, and while Pest is the option you get from the installer if you don't change anything, it is offered more like a choice than a strict default.

> Either way, it's easy to [swap out in Blueprint's config](https://blueprint.laravelshift.com/docs/generated-tests/).

Fair point! The change is not that big indeed.
Thanks for the feedback.

PR #698, opened 2024-09-27
Draft implementation for auto skip functionality

wotta:
Good morning/afternoon/evening @jasonmccreary,

I hope you are doing well. I've made a real basic implementation of how I had the auto skip functionality in mind. Part of the code that fetches the file classes is too basic but it was mainly used to get the point across.
If this is something you don't like at all please share that with me. Either way it up to you if you see some "benefit" in this change.

The one thing I am running into is the test itself. I don't fully understand how the test is working so if you see some benefit in this would you be willing to help me out explaining how the test works. If you don't have time the coming weeks don't hesitate to share that with me.

I will try to start with a test but I cannot guarantee I can make a working test in a short time.

### DIFF
--- a/src/Commands/BuildCommand.php
+++ b/src/Commands/BuildCommand.php
@@ -65,7 +65,7 @@ class BuildCommand extends Command
                 $blueprintContent = $this->filesystem->get($blueprintFile);
                 $blueprintContent = Yaml::parse($blueprintContent);
                 $blueprintClasses = $blueprintContent['created'] ?? [];
-                $skip = implode(',', array_map(function($class) {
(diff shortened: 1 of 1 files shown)


jasonmccreary:
Again, I'm open to it. I need to see a test as I don't really understand how this code works.

Also, an option like `skip-existing` or `ignore-existing` might be more appropriate. I didn't realize we had a `skip` option already.

wotta:
I am going to try to write the tests to show how it is supposed to work. I am struggling to understand the tests logically since it seems to be mocking a lot. Not sure if I mock the logic it will actually run past it.

Issue #703, opened 2024-11-07
Invalid request class name define in controller 


aip-jit:
- Laravel Version:11.30.0
- PHP Version: 8.2.25
- Blueprint Version: ^2.10
- Platform: Linux

### Issue: When I created the request file, it was successfully made with the correct name, but in the controller, the request class name was different. and also different namespace class import in controller  
my generated controller and request file 
### UserController.php:
```
<?php

namespace App\Http\Controllers;

use App\Http\Requests\UserControllerStoreRequest;
use App\Http\Requests\UserControllerUpdateRequest;
use App\Models\User;
class UserController extends Controller
{
    public function store(UserControllerStoreRequest $request): RedirectResponse
    {
        $user = User::create($request->validated());

        return redirect()->route('backend.users.show');
    }
    public function update(UserControllerUpdateRequest $request, User $user): RedirectResponse
    {
        $user->update($request->validated());

        $request->session()->flash('user.id', $user->id);

        return redirect()->route('user.index');
    }
}
```
### UserStoreRequest.php
```
<?php

namespace App\Http\Requests;

use Illuminate\Foundation\Http\FormRequest;

class UserStoreRequest extends FormRequest
{
    /**
     * Determine if the user is authorized to make this request.
     */
    public function authorize(): bool
    {
        return true;
    }

    /**
     * Get the validation rules that apply to the request.
     */
    public function rules(): array
    {
        return [
            'name' => ['required', 'string'],
            'email' => ['required', 'email'],
            'email_verified_at' => ['nullable'],
            'password' => ['required', 'password'],
            'remember_token' => ['nullable', 'string'],
        ];
    }
}
```
### UserUpdateRequest.php
```
<?php

namespace App\Http\Requests;

use Illuminate\Foundation\Http\FormRequest;

class UserUpdateRequest extends FormRequest
{
    /**
     * Determine if the user is authorized to make this request.
     */
    public function authorize(): bool
    {
        return true;
    }

    /**
     * Get the validation rules that apply to the request.
     */
    public function rules(): array
    {
        return [
            'name' => ['required', 'string'],
            'email' => ['required', 'email'],
            'email_verified_at' => ['nullable'],
            'password' => ['required', 'password'],
            'remember_token' => ['nullable', 'string'],
        ];
    }
}
```

### draft.yaml:
```yaml
# PLEASE PASTE THE DRAFT FILE WHICH REPRODUCES THE ISSUE HERE...
controllers:
  UserController:
  store:
[...]

PR #704, opened 2024-11-10
Request and Response class name replacement fixed to handle environment differences

yolcuiskender:
**Request class name is now generated correctly.**
**Response class name is now replaced correctly.**

**PHP_EOL** could cause problems in different environments like based on OS and Platform.

**Example:**
Windows terminal with Apache would evaluate PHP_EOL as \r\n at runtime while the stub only has \n causing the str_replace function to fail.

Furthermore tests use \r\n in the files opposit to \n within the stubs causing false negatives.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .env
 /.phpunit.cache
 composer.lock
+/.idea/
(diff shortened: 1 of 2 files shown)


jasonmccreary:
Thanks. But tests are failing. There is also no test to prove the issue/fix.

Also, no one has reported a problem with `PHP_EOL`. Either way, best to keep your PRs focused to a single "issue".

yolcuiskender:
The failing tests in the pipeline seem to be unrelated to the issue itself, they originate from cache v2 in the workflow for macos, not something I remember changing. This PR solves both #703 and #705.

jasonmccreary:
I'll fix the build on `master`. Either way, it'd still like to see a test for you fix as currently this seems either untested or not an issue.

PR #734, opened 2025-01-07
Inertia page generator

nicodevs:
Given the introduction of the `inertia` statement to indicate that the controller should return `Inertia::render`, it would be great if Blueprint could generate Inertia page components for the appropriate adapter (Vue, React, Svelte), similar to how the `render` statement triggers the generation of Blade templates.

### Changes

This PR implements that feature by introducing an `InertiaPageGenerator`.

- It determines the adapter (Vue, React, Svelte) by reading the dependencies listed in `package.json`.
- If no adapter is found, it will not generate page components.
- If an adapter is detected, it iterates through the controller and statements, generating a page component using a stub appropriate for the adapter.
- It populates the stub with the statement data as props for the page component.

The stubs are fairly minimal and are based on the examples provided in the [Inertia documentation](https://inertiajs.com/pages).

### Example

Given the following draft:

```yaml
controllers:
  Customer:
    index:
      query: all
      inertia: Customer/Index with:customers
    show:
      inertia: Customer/Show with:customer,foo,bar
```

And having this dependency in `package.json`:

```json
{
    "dependencies": {
        "@inertiajs/vue3":"^2.0.0",
        // more dependencies...
    }
}
```

Blueprint will generate these page components:

- `resources/js/Pages/Customer/Index.vue`

```vue
<script setup>
import { Head } from '@inertiajs/vue3'

defineProps(["customers"])
</script>

<template>
  <Head title="Customer Index" />
  <h1>Customer Index</h1>
</template>
```

- `resources/js/Pages/Customer/Show.vue`

```vue
<script setup>
import { Head } from '@inertiajs/vue3'

defineProps(["customer","foo","bar"])
</script>

<template>
  <Head title="Customer Show" />
  <h1>Customer Show</h1>
</template>
```

### Testing

This PR adds tests to check the output for the three adapters (Vue, React, Svelte).

If this doesn’t align with the goals of Blueprint, I can convert it into a plugin. Thank you for taking the time to review it!

### DIFF
--- a/config/blueprint.php
+++ b/config/blueprint.php
@@ -170,6 +170,7 @@ return [
         'notification' => \Blueprint\Generators\Statements\NotificationGenerator::class,
         'resource' => \Blueprint\Generators\Statements\ResourceGenerator::class,
         'view' => \Blueprint\Generators\Statements\ViewGenerator::class,
+        'inertia_page' => \Blueprint\Generators\Statements\InertiaPageGenerator::class,
         'policy' => \Blueprint\Generators\PolicyGenerator::class,
     ],
 
--- a/src/Generators/Statements/InertiaPageGenerator.php
+++ b/src/Generators/Statements/InertiaPageGenerator.php
@@ -2,19 +2,33 @@
 
 namespace Blueprint\Generators\Statements;
 
-use Blueprint\Models\Controller;
 use Blueprint\Contracts\Generator;
 use Blueprint\Generators\StatementGenerator;
 use Blueprint\Models\Statements\InertiaStatement;
 use Blueprint\Tree;
+use Illuminate\Support\Str;
 
 class InertiaPageGenerator extends StatementGenerator implements Generator
 {
-    protected array $types = ['controllers', 'views'];
+    protected array $types = [];
+
+    protected array $adapters = [
+        'vue3' => ['framework' => 'vue', 'extension' => '.vue'],
+        'react' => ['framework' => 'react', 'extension' => '.jsx'],
+        'svelte' => ['framework' => 'svelte', 'extension' => '.svelte'],
+    ];
+
+    protected ?array $adapter = null;
 
     public function output(Tree $tree): array
     {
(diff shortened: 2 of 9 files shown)


config/blueprint.php
```diff
@@ -170,6 +170,7 @@
         'notification' => \Blueprint\Generators\Statements\NotificationGenerator::class,
         'resource' => \Blueprint\Generators\Statements\ResourceGenerator::class,
         'view' => \Blueprint\Generators\Statements\ViewGenerator::class,
+        'inertia_page' => \Blueprint\Generators\Statements\InertiaPageGenerator::class,
```
jasonmccreary, line 173: Is this the official name of these "views". Or could we simply call it `inertia`?

jasonmccreary (commented):
Looks good. But quick question on naming as I don't use Inertia.

config/blueprint.php
nicodevs: This is the official name they use in their docs: https://inertiajs.com/pages#creating-pages

But yeah, nothing prevents us from simply saying:

```suggestion
        'inertia' => \Blueprint\Generators\Statements\InertiaGenerator::class,
```

As you prefer!

jasonmccreary: If that's what they use, then no problem.

nicodevs:
Thank you @jasonmccreary!

mrlinnth:
Since I am using blueprint with laravel 12 react starter kit, I really appreciate this feature. 

May I clarify one thing?

Will this also work for `resource` shorthand? 

Let's say if my yaml is

```yaml
controllers:
    Todo:
        resource: web
```

And I have `"@inertiajs/react": "^2.0.0"` in my package.json file.

Will this generate necessary 7 actions for the controller with Inertia::render and generate necessary jsx or tsx files under resources/js dir?

jasonmccreary:
Currently it will not. But I think that'd be an excellent PR. Probably not sniff out the dependencies as there could be other. But rather a new configuration option for "default view" and allow `blade`, `inertia`, maybe `livewire`.

PR #736, opened 2025-01-12
Use `fake` helper in factories

nicodevs:
This PR updates the generated code to use the `fake()` helper function instead of `$this->fake`, aligning with the latest recommendations in the Factories documentation of [Laravel 10](https://laravel.com/docs/10.x/eloquent-factories) and [Laravel 11](https://laravel.com/docs/11.x/eloquent-factories).

No new features have been introduced, and all existing tests are passing.

### How to Test Locally

1. Install Laravel.

```bash
composer create-project laravel/laravel example-app
```

2. Install this version of the Blueprint package.

3. Create a standard `draft.yaml`

```yaml
models:
  Post:
    title: string
controllers:
  Post:
    resource: web
```

4. Generate:

```bash
php artisan blueprint:build
```

**Expected:** The generated factory should now utilize the `fake()` helper instead of `$this->fake`. Migrations should run successfully without errors.


jasonmccreary:
Where does Laravel **recommend** the `fake` helper?

nicodevs:
@jasonmccreary ~ Sorry if that was not the appropriate term; perhaps I should have said "aligning with the latest **examples** in the Factories documentation of Laravel 10 and Laravel 11"

Taylor made the change (`$this->fake` to `fake()`) to the default User factory in 2022. Here is the link to the commit: https://github.com/laravel/laravel/commit/0c3d1fabe5a9319131aa4e9d74b27f29e74a50fb

Given this context, I think users might expect `fake()` to be used in factories. But feel free to disregard this suggestion if not!

jasonmccreary:
Yes, I see it in the docs as the example. I'm not against it. Just may think on it a bit to ensure this is the new "convention".

PR #746, opened 2025-04-14
Add support for adding custom traits to models

deegital:
I had the need to add additional traits to the models, in my case it was spatie's laravel-permission HasRules trait. This PR adds compatibility for the following syntax:

```yaml
User:
    name: string
    traits:
      - Spatie\Permission\Traits\HasRoles
```


jasonmccreary:
This is probably not something I'd add myself. However, since you have done the work, I am willing to if you can make some small changes:

- [ ] Move `traits` under the existing `meta` property to minimize naming conflicts.
- [ ] Ensure this is supported for controllers as well.

deegital:
Sounds good, I will add those changes!

PR #748, opened 2025-04-18
Add typing to HasFactory the Laravel way

mondesirm:
Hi, looks like I was beat to it regarding this [commit](https://github.com/laravel-shift/blueprint/commit/1a0a5bbe78af3af8e18ea667c359307f8eef915f).

Context: I'm currently working on a project that uses Laravel's React Starter Kit and I found myself having to customize some stubs and generators more than I thought I would.

Like your last commit and this one, I will create some pull requests that aligns more with Laravel's new starter kits and the way `php artisan make:model` creates files.

### DIFF
--- a/src/Generators/ModelGenerator.php
+++ b/src/Generators/ModelGenerator.php
@@ -462,7 +462,7 @@ class ModelGenerator extends AbstractClassGenerator implements Generator
 
         sort($traits);
 
-        return Str::replaceFirst('use HasFactory', 'use ' . implode(', ', $traits), $stub);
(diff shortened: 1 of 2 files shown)


mondesirm:
When I cloned the repo and ran `pint`, it told me that 101 files had the wrong "line_ending" and when I checked them, they were all using CRLF and now use LF. Is it normal?

I had to revert the changes and just focus on my pull request or else I'd have a messy commit.

jasonmccreary:
Closing this for now. Devs can update the stubs if they want them customized. If you really believe this should be core, then I'd rather see it added when the docblock config option is set.

PR #755, opened 2025-10-01
Fix MigrationGenerator for some columns with underscore

nikdejan:
Before this fix for columns like this:
agent_party_id: id foreing:agents.party_id
migration was generated like this:
$table->foreignId('agent_party_id')->constraints(); Which is a problem because this syntax assumes table name is 'agent_parties'.

### DIFF
--- a/tests/Feature/Generators/MigrationGeneratorTest.php
+++ b/tests/Feature/Generators/MigrationGeneratorTest.php
@@ -613,6 +613,43 @@ final class MigrationGeneratorTest extends TestCase
         $this->assertEquals(['created' => [$user_migration, $team_migration, $pivot_migration]], $this->subject->output($tree));
     }
 
+    #[Test]
+    public function output_creates_constraint_for_foreign_key_to_non_id_column_with_underscore(): void
+    {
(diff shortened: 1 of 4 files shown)


jasonmccreary:
Thanks for this. I would appreciate seeing a test fixture that drove out the solution or document/verifies the change.

nikdejan:
I agree. What you just described: `agent_party_id: id foreign:agents` is a behavior I expected when I first time started using blueprint. It should be a feature.
But BUG I was talking about is different thing, although related to what you said. 
Docs:
`If you are not following Laravel's naming conventions, you may set the attribute on the foreign modifier. Blueprint supports either the foreign table name or THE TABLE NAME AND COLUMN name using dot notation.` 
So, when not following conventions, and using custom table name and column, as per docs, if column name has an underscore and if accidentally `foreign key = concat(table, primary_key)` migration is not generated correctly.

jasonmccreary:
I agree there is a bug here. My point is that this PR only contains part of the solution. Maybe that's fine. But I'd personally like to have the other items fixed too.

Issue #756, opened 2025-10-19
Adding attribute name for BelongsToMany sometimes causes migration error


LorenzoPapi:
- Laravel Version: 12.34.0
- PHP Version: 8.4.13
- Blueprint Version: 2.12.0
- Platform: Linux

### Issue:
I have a many to many relationship between AppRoles and AppUsers: each user HasMany roles and each role BelongsToMany users. Both the models use ulids: however, adding `users` after the line `belongsToMany: AppUser` causes the Blueprint to generate a migration where, instead of the usual `foreignUlid`, a `foreignId` is generated. If I remove the name it works flawlessly. I also have foreign key constraints active.

### draft.yaml:
```yaml
models:
  AppRole:
    ulid

    slug: string:20:unique
    name: string:50

    relationships:
      belongsToMany: AppUser:users
  
  AppUser:
    ulid
    
    username: string:255:unique
    password: password
    firstName: string:255
    lastName: string:255
    studies: string:255 nullable
    room: string:20 nullable
    telephone: string:20 nullable
    description: longtext
    
    softDeletes
    timestamps: false

    relationships:
      hasMany: AppRole:role
      belongsToMany: Group:group

  Group:
    name: string
    color: string

    relationships:
      hasMany: AppUser:user
```

LorenzoPapi:
Alright, I'll try to open it today. Also, I added the ability to insert traits as a meta attribute, at least for models because that's what I needed. If I add it to controllers too and open another PR would it get merged?

jasonmccreary:
Sure. Please make them separate PRs though. Also, double check adding traits as meta data doesn't already rexist. I felt like it did. But maybe that was `extends` only.

jasonmccreary:
After looking at this more closely, setting `meta` when you use an unconventional table name is the appropriate solution, and works currently. Generally speaking, there should be no reason to set the table name when providing a relationship. Especially a conventional one.

Feel free to open a separate PR for the `extends`/`use` `meta` options.

PR #757, opened 2025-10-20
Fix pivot table migration building

LorenzoPapi:
When the belongstomany attribute was given a special name, the id column type identifying wasn't working and would fall back to the usual integer id, causing errors. Now, by considering as a key the string before the `:`, it should work.


jasonmccreary:
Thank you for your PR. However, once I added the appropriate `meta` statement for the custom table name (e.g. `users`) to your sample `draft.yaml` file, the migrations were generated with `foreignUlid()` correctly. As such, I'm closing this and Issue #756.

Issue #764, opened 2025-11-23
Incorrect Path Generated for nested Resources and Requests


khyamshahzad:
- Laravel Version: 12.0
- PHP Version: 8.4
- Blueprint Version: 2.13
- Platform:  Linux

### Issue:
While generating a nested controller with API resource, the path for associated Resource and Request files is getting corrupted. 

The path for example appears as "_app/Http/Resources/**Api\V1**/MasterRecordOptionResource.php_" instead of "_app/Http/Resources/**Api/V1**/MasterRecordOptionResource.php_". 

### draft.yaml:
```yaml
models:
  MasterRecord:
    name: string
    short_name: string unique
    relationships:
      hasMany: MasterRecordOption
  MasterRecordOption:
    master_record_id: id foreign:master_records.id
    option_text: string
    relationships:
      belongsTo: MasterRecord

controllers:
  Api\V1\MasterRecordController:
    resource: api
  Api\V1\MasterRecordOptionController:
    resource: api

```
### Solution:

Update the _function getStatementPath_  in _src/Generators/Statements/ResourceGenerator.php_ and in _src/Generators/Statements/FormRequestGenerator.php_ as below.

**ResourceGenerator.php**

`
    protected function getStatementPath(string $name): string
    {
        return str_replace('\\', '/', Blueprint::appPath() . '/Http/Resources/' . $name . '.php');
    }
`

**FormRequestGenerator.php**

`
    protected function getStatementPath(Controller $controller, string $name): string
    {
        return str_replace('\\', '/', Blueprint::appPath() . '/Http/Requests/' . ($controller->namespace() ? $controller->namespace() . '/' : '') . $name . '.php');
    }
`

jasonmccreary:
I see. Feel free to open a PR. Otherwise, I'll try to get to this in my next Blueprint livestream (probably late December).

PR #765, opened 2025-11-24
Fixes incorrect path generated for nested Resources and Requests directories.

khyamshahzad:
Resolves the issue # 764 where directory path for nested resources and requests folders were getting corrupted. 
https://github.com/laravel-shift/blueprint/issues/764


khyamshahzad:
Here is sample .yaml file that was failing with generating the nested directory structure due to incorrect path separator. 

```
models:
  MasterRecord:
    name: string
    short_name: string unique

controllers:
  Api\V1\MasterRecordController:
    resource: api
```

PR #774, opened 2026-07-31
Create hasMany relationships in resource stores

richardrh:
## Summary

Allow an API resource `store` action to create its model and one declared
`hasMany` relationship in the same request and database transaction.

Current nested resources create one child after its parent already exists:

```http
POST /api/orders/{order}/items
```

Blueprint infers the request payload key from the related model (`items`) and resolves it to the conventional `items()` hasMany relation declared on `Order`.

This change supports creating an Order and its Items together:

```http
POST /api/orders
```

## Syntax

```yaml
controllers:
  Api/Order:
    resource: api
    meta:
      store: Item
```

## Request JSON

```json
{
  "reference": "ORDER-123",
  "items": [
    {
      "product_id": 1,
      "quantity": 2
    },
    {
      "product_id": 2,
      "quantity": 1
    }
  ]
}
```

## Generated Controller example
```php
public function store(OrderStoreRequest $request): OrderResource
{
    $order = DB::transaction(function () use ($request) {
        $order = Order::create(
            $request->safe()->only(['reference'])
        );

        $order->items()->createMany(
            array_map(
                fn (array $item) => Arr::only(
                    $item,
                    ['product_id', 'quantity']
                ),
                $request->validated('items')
            )
        );

        $order->load(['items']);

        return $order;
    });

    return new OrderResource($order);
}
```

## Scope

- Resource `store` actions only
- One child model
- One level deep
- Conventional declared `hasMany` relationships
- Aliased relationships are rejected
- No nested update, synchronization, deletion, `hasOne`, or `belongsToMany`

## Validation

- Pint passes
- Focused Blueprint tests pass
- Generated PHPUnit and Pest tests pass in fresh Laravel 13 applications
- Runtime validation created one Order and two Items with matching foreign keys


jasonmccreary:
@richardrh, I'm not familiar with aggregate API endpoints - what is the real-world use case for this feature?

Also, if this is only `store` endpoints and relationships, we could probably tighten up the `meta`. Maybe:

```yaml
meta:
   aggregate: Item
```

richardrh:
I agree it isn't worded well. it is simply creating multiple items for a parent relationship in one transaction. so Orders -> hasMany to Items the user can submit one single post request. With the current nested resources client has to submit one order with one item.

For the confusion on what aggregate endpoints mean and also taking your better suggestion on how the meta: key would work, we could do something like this, which seems to intuitively tell the user that the Order controller will store multiple Items.

And alternative to below might be meta: createMany: Item which is close to the Laravel syntax.

@jasonmccreary  would you prefer one of these?

```yaml
    meta:
      store: Item
```

```yaml
meta:
  createMany: Item
```

jasonmccreary:
@richardrh, whatever aligns with Laravel and is the least to type. That's what aligns with Blueprint's goals.

If you make those adjustments and ensure all tests are passing, I'm willing to merge.

richardrh:
Jason, updated to the shorter Laravel resource-action syntax `meta.store: Item` and changed the main builder function to `buildStoreWithRelation` which matches the syntax.

jasonmccreary:
Looks like some errors. Not sure if they are related.

richardrh:
I traced the three failures to the existing `EraseCommandTest` expectation: Symfony YAML 8.1 writes an empty map as `{}` while the mock expects the older `{  }` whitespace.

Separare PR to fix the CI in #775

jasonmccreary:
@richardrh, thanks I merged #775. So if you want merge or rebase this, I assume we'll be all passing again.

richardrh:
update with a new CI run

